PR #111, opened 2014-12-15
- mbuffer timeout: 60 sec

hadfl:


smaugs:
Hi,

the mbuffer "-W" option is not present on ubuntu 14.04 neither on debian jessie:

mbuffer -W
mbuffer: fatal: unknown option "-W"

I deleted the -W option in my source - maybe it's possible to implement a check for the existence of the "-W" before using it in znapzend.

smaugs:
PS: to suppress the warnings about the console and the autoloader it would be cool to add

-A /bin/false

to the mbuffer switches.

PR #129, opened 2015-01-17
fix configure path tests

wiedi:
Without this change it is not possible to use configure --mandir=...


oetiker:
Hmmm the only point of this line is to make the change possible. the default content of mandir is actually '${datarootdir}/man' have a look at the generate configure script

oetiker:
so there was a space missing ... 4cdbcde and 890a750 fix it

wiedi:
thanks!

Issue #131, opened 2015-01-18
Install with brew on OSX


lenada:
It would be very convenient for znapzend users on osx to use brew for installation and updates.
something like the following homebrew Formula `znapzend.rb` needs to be shipped to @homebrew.

https://gist.github.com/lenada/e773a65aa9d37c0eb42c

@oetiker what is your general take on all those package managers out there? would you be willing to maintain and update some, for example the homebrew formula, linux apt, etc. yourself? or are you rather looking for external maintainers.
should I PR my Formula against https://github.com/Homebrew/homebrew ? who should do so on the next release?

by the way, znapzend is a really great tool  :) appreciated!

lenada:
I also would be really interested in getting znapzend into freebsd ports :neckbeard:

oetiker:
@lenada OS are certainly welcome ... I had a look at your homebrew gist. It would be preferable if it worked from the standard source version though and not from the pre-built one ... the pre-build is really a hack ...

lenada:
@oetiker I gave the source version a try first but failed due to `make get-thirdparty-modules` (also related to #125).

also symlinks to executables do not work because of `$FindBin::Bin/../lib` & `$FindBin::Bin/../thirdparty`

oetiker:
but the source version and the pre-built are the same, except that get-thirdparty-modules has been executed prior to make install ...

in anyway, it might be a good idea to replace the FindBin magic upon installation using the proper path

lenada:
@oetiker problem is ./configure --prefix=/usr/local/Cellar/znapzend/0.14.0 fails on first run due to missing thirdparty dependencies, which causes the build to fail. I probably could try to enforce/override the exit status, but that seems even more hackish.

IMHO it would be nice to have a configure flag to skip the thirdparty check. That way I could install the perl/mojo dependencies through brew and get rid of the thirdparty directory alltogether.

`/usr/bin/gnumake get-thirdparty-modules` also has a dependency on `wget` which is not installed on os x per default.

oetiker:
would curl be better ?

```
./configure --prefix=/usr/local/Cellar/znapzend/0.14.0 || true
gnumake get-thirdparty-modules
```

lenada:
@oetiker for os x curl is preinstalled. But wget vs curl etc. may vary across operating systems / distros?

oetiker:
configure could check

oetiker:
[...]

Issue #125, opened 2015-01-17
install should not create thridparty directory in prefix directory


diekhans:
thirdparty software install creates a directory ${prefix}/thirdparty.  This violates unix/linux filesystem
organization rules (seee man hier).  These should probably go in ${prefix}/lib/znapzend/thirdparty

oetiker:
the idea is to install in /opt/znapzend-$version in that case all is fine. but I agree if you want to throw it into /usr then this does not work well ...

dewi-ny-je:
Be careful with zones: /usr/local is inherited and /opt is not
http://compgroups.net/comp.unix.solaris/-usr-local/354465

Issue #120, opened 2015-01-01
Doesn't run on FreeBSD 9.3 


cedrictineo:
Hello, I tried to install Znapzend 0.14.0 following both the binary installation and the compile installation process (including the make get-thirdparty-modules step) - as per the readme file and in both cases when trying to run /opt/znapzend-0.14.0/bin/znapzendzetup I get the following error message :

root@abch026:/opt/znapzend-0.14.0/bin # /opt/znapzend-0.14.0/bin/znapzendzetup
Can't locate Mojo/Base.pm in @INC (you may need to install the Mojo::Base module) (@INC contains: /opt/znapzend-0.14.0/bin/../thirdparty/lib/perl5 /opt/znapzend-0.14.0/bin/../lib /usr/local/lib/perl5/site_perl/mach/5.18 /usr/local/lib/perl5/site_perl /usr/local/lib/perl5/5.18/mach /usr/local/lib/perl5/5.18 /usr/local/lib/perl5/site_perl/5.18 /usr/local/lib/perl5/site_perl/5.18/mach .) at /opt/znapzend-0.14.0/bin/znapzendzetup line 12.
BEGIN failed--compilation aborted at /opt/znapzend-0.14.0/bin/znapzendzetup line 12.

root@abch026:/opt/znapzend-0.14.0/bin # perl --version
This is perl 5, version 18, subversion 4 (v5.18.4) built for amd64-freebsd-thread-multi

I haven't found a way to resolve that on my own. Can someone help?

Thank you and have a great and Happy New Year !

Cedric Tineo

cedrictineo:
Ok - after trying some more - here's what got me to get it up and running on FreeBSD 9.3 :
(assuming a FreeBSD 9.3 with pkg installed)

pkg install wget
pkg install perl5
pkg install p5-mojo
wget --no-check-certificate https://github.com/oetiker/znapzend/releases/download/v0.14.0/znapzend-prebuilt-0.14.0.tar.gz
#the --no-check-certificate was required to work
tar zxvf znapzend-prebuilt-0.14.0.tar.gz
cp -rp znapzend-prebuilt-0.14.0 /opt/znapzend-0.14.0

then adding the path /opt/znapzend-0.14.0/bin to the set path directive in .cshrc

oetiker:
did you, by any chance neglect to run the configure script and then the dependency build script ?

cedrictineo:
[...]

PR #132, opened 2015-01-21
Be good

oetiker:
* no thirdparty dir in install tree
* build missing perl modules automatically
* no FindBin in installed version of the program


oetiker:
Maybe fixes #131, #125, #120

PR #139, opened 2015-03-16
- fix 'zfs list' if listsnapshots zpool property is on

hadfl:


lib/ZnapZend/ZFS.pm
```diff
@@ -125,6 +125,7 @@ sub snapshotExists {
 sub listDataSets {
     my $self = shift;
     my $remote = shift;
+    my @dataSets;
 
     my @ssh = $self->$buildRemote($remote, [@{$self->priv}, qw(zfs list -H -o name)]);
```
oetiker: how about calling

````zfs list -t filesystem -H -o name````

hadfl: oh, my bad, thought `-t` only works w/ `snapshot|volume`

will patch the patch :D

hadfl: `-t filesystem` is not an option as zvols would be neglected in that case...

PR #140, opened 2015-03-16
- use option '-t filesystem' to get datasets

hadfl:


hadfl:
nuked, because zvols would be ignored, too

PR #147, opened 2015-05-16
- allow '.' and '-' for zfs user properties

hadfl:


bin/znapzendzetup
```diff
@@ -61,7 +61,7 @@ sub parseArguments {
             next;
         };
         #keyword DST detected. start setting up destination filesystem and plan
-        /^DST(?::([a-z0-9_]+))?/ && do {
+        /^DST(?::([-\.\w]+))?/ && do {
```
oetiker: \w is locale sensitive, this means the result will differ according the current locale, I don't think that is ideal.

how about being just a little more accomodating?

`[-_.a-zA-Z0-9]+`

Note, a `.` does not have to be escaped here.

PR #163, opened 2015-08-23
Include note on lower-case requirement for DST fields

JMoVS:
- Added a little sentence to clarify that DST names can only contain lower-case letters


oetiker:
thanks

PR #174, opened 2015-11-20
added script to create packages

morph027:
We've got more than one host, so creating a package and pushing to our internal repo is much easier than building on every host.


oetiker:
Hi @morph027, since there are many packaging systems around, I guess it would be wise to structure this in a way that not only yours works. Maybe a subdirectory packaging/checklinstall and in there your script and the relevant part of the README.md

morph027:
Jepp, sounds great. We could also add fpm as creator in a second step, as it can handle way more package formats (but requires ruby...)

oetiker:
Hi @morph027 almost there ... if you can move the documentation into the checkinstall directory and just add a note to the main readme file pointing to the packaging directory that would be great ... I would like to make sure that new users get confused as little when reading the readme as possible.

oetiker:
thanks

PR #175, opened 2015-11-24
Fix sudo/pfexec use when querying config.

doctorjames:
`refreshBackupPlans` calls `$self->zConfig->getBackupSetEnabled($dataSet)`
but `zConfig` was not initialised with sudo/pfexec features and hence
there were errors calling the `zfs` command.


oetiker:
thanks

PR #176, opened 2015-11-24
Fix sudo/pfexec use when initialising ZnapZend::Config.

doctorjames:
Same issue as  #175 only in stats command this time.


oetiker:
thanks

PR #180, opened 2015-11-30
added systemd unit file

morph027:
tested 2 weeks now on debian 8


oetiker:
since this file does not get used automatically by configure/make could you add a comment at the top of the file explaining it's purpose ?

oetiker:
thanks

PR #184, opened 2016-01-16
Comparing with == is a bashism

wiedi:
Comparing with one = is enough and fixes the build in POSIX shells that are not bash.


oetiker:
thanks

PR #189, opened 2016-02-07
Fix a bug (Alias) in the systemd service file, and improve the documentation

antoneliasson:
See the commit messages for justification. I'm happy to provide further clarification if needed.


antoneliasson:
Thanks for the quick response!

Issue #168, opened 2015-09-22
ZnapZend.pm respect system signals (SIGINT, SIGTERM, SIGHUP) only when executing a plan


chriseqipe:
On debian wheezy, when sending SIGTERM or SIGINT to the perl process, znapsend.pm only checks for signals on the next run. If for example the plan takes a snapshot every 10 minutes, it takes maximum 10 minutes to end the process.  Maybe you should start a faster worker listening every minute to system signals.

oetiker:
What would you like znapzend do when it receives a signal ? abort in mid operation ? for that you can send it a SIGKILL.

martin-rueegg:
Thank you, @chriseqipe for pointing out this issue. I've come across it when writing an init.d script and wanted to _cleanly_ shutdown `znapzend`, which is probably what you wanted as well. My config was set to take one snapshot per hour, and it was quite frustrating that `znapzend` did not react on any signal (SIGTERM, SIGINT, SIGHUP), and the process could only be SIGKILLed (which is not performed by the script, but by the interpreter). After spending a day or so looking for the reason, i came across your issue and it helped me to track the cause.
 
unfortunately, @oetiker did not understand your post (of course, we _don't_ want to use SIGKILL, unless absolutely required!). so I'll try to explain it in other words:

The following is a diagram of the activity and events over time:
```
────> time ────>────>────>─────>─────>─────>────>────>──────>

signal─>─┬───────────────────────┬────────────────────┐
 ┌───────┴──┐                  ┌─┴─┐                ┌─┴─┐
[#######@@@@@]---------------*[@@@@@]-------------*[@@@@@]...
 ╘══╤══╛╘═╤═╛ ╘══════╤══════╛│ ╘═╤═╛ ╘═════╤═════╛│ ╘═╤═╛
   (1)   (2)        (3)     (4) (2)       (3)    (4) (2)
    │     │          │       │
    │     │          │       └─> (4) timer fires
    │     │          └─────────> (3) wait for next timer to fire
    │     │                          (-> which is not _sleep_!)
    │     └────────────────────> (2) Worker running
    └──────────────────────────> (1) startup
```
Signals will only be received and handled by `znapzend` during time of activity (between the `[brackets]`), as the program has no running code during the times where only timers are scheduled (`------`).

[...]

PR #190, opened 2016-02-23
add keepalive to receive signals at all times.

martin-rueegg:
solves https://github.com/oetiker/znapzend/issues/168

in order to avoid an error message during shutdown, the following PR should be applied https://github.com/jberger/Mojo-IOLoop-ForkCall/pull/9.

### DIFF
--- a/lib/ZnapZend.pm
+++ b/lib/ZnapZend.pm
@@ -113,6 +113,7 @@ my $killThemAll  = sub {
         waitpid($backupSet->{send_pid}, WNOHANG)
             || kill(SIGKILL, $backupSet->{send_pid}) if $backupSet->{send_pid};
     }
+    $self->zLog->info('Done.');
     exit 0;
 };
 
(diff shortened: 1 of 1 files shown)


oetiker:
Your patch breaks at least the test suite ... as far as I understand, the reporter of #168 takes issue with the fact that znapzend does not abort the current zfs send operation ... or am I missing something ?

martin-rueegg:
i don't know how my patch would break [`zfs recv`](https://travis-ci.org/oetiker/znapzend/jobs/111128736#L843) and [`zfs send`](https://travis-ci.org/oetiker/znapzend/jobs/111128736#L844).

is that really on a network connection? could there be an issue with the infrastructure? i don't think the worker threads should be blocked by the keepalive thread. that's the whole point about the threads, ain't it?
in order to test it here, could i send it to the same host using ssh? (i don't currently have an other host with a zfs filesystem)

oetiker:
znapzend contains mock ssh and zfs commands ...

martin-rueegg:
close due to https://github.com/oetiker/znapzend/pull/191#issuecomment-187680508

PR #191, opened 2016-02-23
additional [info] and [debug] messages

martin-rueegg:
an alternate PR for https://github.com/oetiker/znapzend/pull/190 in order to see if the recurring method would not break the test suite.
would also fix https://github.com/oetiker/znapzend/issues/168
discussion after build-check ..

### DIFF
--- a/lib/ZnapZend.pm
+++ b/lib/ZnapZend.pm
@@ -96,7 +96,7 @@ has zLog => sub {
 my $killThemAll  = sub {
     my $self = shift;
 
-    $self->zLog->info('terminating znapzend...');
+    $self->zLog->info("terminating znapzend (PID=$$) ...");
     #set termination flag
     $self->terminate(1);
 
@@ -113,6 +113,8 @@ my $killThemAll  = sub {
         waitpid($backupSet->{send_pid}, WNOHANG)
             || kill(SIGKILL, $backupSet->{send_pid}) if $backupSet->{send_pid};
     }
+
+    $self->zLog->info("znapzend (PID=$$) terminated.");
     exit 0;
 };
 
@@ -523,12 +525,16 @@ my $daemonize = sub {
 sub start {
     my $self = shift;
 
+    $self->zLog->info("znapzend (PID=$$) starting up ...");
+
     $self->$daemonize if $self->daemonize;
 
     # set signal handlers
-    $SIG{INT}  = sub { $self->$killThemAll; };
-    $SIG{TERM} = sub { $self->$killThemAll; };
+    $SIG{INT}  = sub { $self->zLog->debug('SIGINT received.'); $self->$killThemAll; };
+    $SIG{TERM} = sub { $self->zLog->debug('SIGTERM received.'); $self->$killThemAll; };
     $SIG{HUP}  = sub {
+        $self->zLog->debug('SIGHUP received.');
+
         #remove active timers from ioloop
         for my $backupSet (@{$self->backupSets}){
             Mojo::IOLoop->remove($backupSet->{timer_id}) if $backupSet->{timer_id};
@@ -540,6 +546,8 @@ sub start {
     $self->$refreshBackupPlans($self->runonce);
 
     $self->$createWorkers;
+
+    $self->zLog->info("znapzend (PID=$$) initialized -- resuming normal operations.");
 
     #start eventloop
     Mojo::IOLoop->start;


hadfl:
if this is really necessary that signals work on any system i would prefer a simple `Mojo::IOLoop->recurring(1 => sub { });`

lib/ZnapZend.pm
```diff
@@ -520,15 +521,31 @@ my $daemonize = sub {
...
 };
 
+my $keepalive = sub {
+    my $self = shift;
+
+    $self->zLog->debug('Keepalive started.');
+
+    # workaround for https://github.com/oetiker/znapzend/issues/168
+    my $heartbeat = Mojo::IOLoop->recurring(1 => sub { my $loop = shift; select(undef, undef, undef, 0.9); });
+
```
oetiker: this is actually pretty evil ... we are in an event loop here and you are sleeping away for 0.9 seconds every second ... the whole point of writing event based programs is that all your tasks should return control to the event loop as quickly as possible ...sleeping is sort of the ultimate mockery ...

martin-rueegg:
@oetiker, yes, i have now implemented the version suggested by @hadfl.

however i guess the check will fail again ...

oetiker:
see, it works. but as I said, your version check is a red herring. the problem was that you tried to run it with mojo 4.x !

martin-rueegg:
> the problem was that you tried to run it with mojo 4.x !
yes, as it seems, i did not re-run the `configure` script! shame on me.

so you think the workaround is obsolete?

would you still consider to implement the additional debug messages?

oetiker:
and you did not run make install ... and yes, extra debug/info messages are fine. except for the 'done'. messages should always have context ... as they are read one line at a time.

martin-rueegg:
> and you did not run make install ... 

actually, i did run `make install`, but without `./configure`ing again.

ok, cool. so i will
1. remove the keepalive?
1. edit the "Done." into "znapzend (PID $$) terminated."?
1 update the PR title

hadfl:
> would you still consider to implement the additional debug messages?

i don't think that additional debug messages are necessary.
- on HUP znapzend will log that it is going to `refresh backup plans...`
- on INT/TERM znapzend will log that it is `terminating znapzend...`

i don't think that spamming syslog is a good idea.

martin-rueegg:
>  i don't think that spamming syslog is a good idea.

agree. but they are [debug] messages. and on production, loglevel should be set to something else than the default [debug]. maybe this default should be reconsidered.

however i do not insist in those debug messages.

oetiker:
update your patch :) debug messages are fine ... its just that if there is already a termination message, you might want to 'enrich' it instead of adding a second one ...

martin-rueegg:
> update your patch :) 

i will.

> its just that if there is already a termination message, you might want to 'enrich' it instead of adding a second one ...

well, actually to me the there are four events that seem to be quite important for system survillance:
1. process initialization (did it get the start command?)
2. process initialization complete (lots of programs do so. it gives me a peace of mind, that initialization worked. it's what i'm looking for when watching syslog!)
3. process shutting down. (did it react to my `service stop` request?)
4. process terminated. (did it really finish cleanup or is it still hanging?)

i would strongly recommend to send all these four messages as [info]!

oetiker:
makes sense

martin-rueegg:
```
[Tue Feb 23 17:01:56 2016] [info] znapzend (PID=15071) starting up ...
[Tue Feb 23 18:18:29 2016] [info] znapzend (PID=19694) starting up ...
[Tue Feb 23 18:18:29 2016] [info] refreshing backup plans...
[Tue Feb 23 18:18:30 2016] [info] found a valid backup plan for rpool/ROOT...
[Tue Feb 23 18:18:30 2016] [info] found a valid backup plan for store...
[Tue Feb 23 18:18:30 2016] [info] znapzend (PID=19694) initialized -- resuming normal operations.

[Tue Feb 23 18:18:41 2016] [debug] SIGHUP received.
[Tue Feb 23 18:18:41 2016] [info] refreshing backup plans...
[Tue Feb 23 18:18:41 2016] [info] found a valid backup plan for rpool/ROOT...
[Tue Feb 23 18:18:41 2016] [info] found a valid backup plan for store...

[Tue Feb 23 18:18:52 2016] [debug] SIGTERM received.
[Tue Feb 23 18:18:52 2016] [info] terminating znapzend (PID=19694) ...
[Tue Feb 23 18:18:53 2016] [info] znapzend (PID=19694) terminated.
```

oetiker:
thanks!

PR #193, opened 2016-02-24
fix bashism introduced in thirdparty makefile (comparing with ==)

wiedi:
fixes the build in POSIX shells that are not bash


oetiker:
thanks :)

PR #194, opened 2016-03-12
Create datasets on dst when it doesn't exist

Wouter0100:
It checks f the `$dstDataSet` exists, if not, it attempts to create it. When creation is succes, the script continues. When it fails, the script stops and continues to the next dataset.

A new `--autoCreation` option is introduced to enable this feature. It's behavior is disabled by default.

### DIFF
--- a/bin/znapzend
+++ b/bin/znapzend
@@ -67,6 +67,7 @@ B<znapzend> [I<options>...]
  --runonce=x        run one round on source dataset x
  --features=x       comma separated list of features to be enabled
  --connectTimeout=x sets the ConnectTimeout for ssh commands
+ --autoCreation     automatically create dataset on dest if it not exists
 
 =head1 DESCRIPTION
 
@@ -167,6 +168,10 @@ use 'sudo' for zfs commands (e.g. when running znapzend as non-priviledged user)
 =item B<--connectTimeout>=I<timeout>
 
 sets the ssh connection timeout (in seconds)
+
+=item B<--autoCreation>
+
+Automatically create a dataset on a destination host if it's not there yet.
 
 =back
 
--- a/bin/znapzend
+++ b/bin/znapzend
@@ -13,7 +13,7 @@ my %featureMap = map { $_ => 1 } qw(oracleMode recvu pfexec sudo);
 
 sub main {
     GetOptions($opts, qw(help|h man debug|d noaction|n nodestroy features=s),
-        qw(daemonize pidfile=s logto=s loglevel=s runonce=s connectTimeout=s)) or exit 1;
+        qw(daemonize pidfile=s logto=s loglevel=s runonce=s connectTimeout=s autoCreation)) or exit 1;
 
     #split all features into individual options
     if ($opts->{features}){
--- a/lib/ZnapZend/ZFS.pm
+++ b/lib/ZnapZend/ZFS.pm
@@ -653,6 +653,10 @@ checks if the dataset exists on localhost or a remote host
 
 checks if the snapshot exists on localhost or a remote host
 
+=head2 createDataSet
+
+creates a dataset on localhost or a remote host
+
 =head2 listDataSets
 
 lists datasets on (remote-)host
--- a/lib/ZnapZend/ZFS.pm
+++ b/lib/ZnapZend/ZFS.pm
@@ -314,9 +314,13 @@ sub sendRecvSnapshots {
     #attemp a creation of the dataset when it doesn't exist
     $dstDataSetExists = $self->createDataSet($dstDataSet) if $self->autoCreation && !$dstDataSetExists;
 
+    ($remote, $dstDataSet) = $splitHostDataSet->($dstDataSet);
+
     #check if the dstDataSet exist on the destination (maybe after a creation)
     !$dstDataSetExists
-        and Mojo::Exception->throw('ERROR: dstDataSet does not exist on dest host');
+        and Mojo::Exception->throw("ERROR: dataset ($dstDataSet) does not exist on "
+	    . "destination ($remote), use --autoCreation to "
+	    . "let ZnapZend auto create datasets");
 
     my ($lastSnapshot, $lastCommon)
         = $self->lastAndCommonSnapshots($srcDataSet, $dstDataSet, $snapFilter);
@@ -331,7 +335,6 @@ sub sendRecvSnapshots {
             . "found on source and destination "
             . "clean up destination $dstDataSet (i.e. destroy existing snapshots)");
 
-    ($remote, $dstDataSet) = $splitHostDataSet->($dstDataSet);
     ($mbuffer, $mbufferPort) = split /:/, $mbuffer, 2;
 
     my @cmd;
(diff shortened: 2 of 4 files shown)


lib/ZnapZend/ZFS.pm
```diff
@@ -122,6 +123,30 @@ sub snapshotExists {
...
 }
 
+sub createDataSet {
+    my $self = shift;
+    my $dataSet = shift;
+    my $remote;
+
+    #just in case if someone aks to check '';
+    return 0 if !$dataSet;
+
```
oetiker: if someone ask us todo something stupid, we should shout bloody murder and not just ignore it

Wouter0100: If copied that one from the other functions, so you want to throw an Exception? :+1:

lib/ZnapZend/ZFS.pm
```diff
@@ -122,6 +123,30 @@ sub snapshotExists {
...
+    ($remote, $dataSet) = $splitHostDataSet->($dataSet);
+    my @ssh = $self->$buildRemote($remote,
+        [@{$self->priv}, qw(zfs create), $dataSet]);
+
+    print STDERR '# ' . join(' ', @ssh) . "\n" if $self->debug;
+
```
oetiker: creating a dataset might be worth a propper logmessage on debug level at least

lib/ZnapZend/ZFS.pm
oetiker: if znapzend asks znapzend todo something stupid it should dy a horrible dath, crying bloody morder while doing so

lib/ZnapZend/ZFS.pm
```diff
@@ -122,6 +123,30 @@ sub snapshotExists {
...
+    #return if 'noaction' or snapshot creation successful
+    return 1 if $self->noaction || !system(@ssh);
+
+    #check if snapshot already exists and therefore creation failed
+    return 0 if $self->dataSetExists($dataSet);
+
```
oetiker: the code below seems to imply that we only try to create the dataset if it does not exist ... sooo

lib/ZnapZend/ZFS.pm
```diff
@@ -289,6 +314,15 @@ sub sendRecvSnapshots {
...
     my $mbufferPort;
+    my $dstDataSetExists = $self->dataSetExists($dstDataSet);
+
+    #attemp a creation of the dataset when it doesn't exist
+    $dstDataSetExists = $self->createDataSet($dstDataSet) if $self->autoCreation && !$dstDataSetExists;
+
+    #check if the dstDataSet exist on the destination (maybe after a creation)
+    !$dstDataSetExists
+        and Mojo::Exception->throw('ERROR: dstDataSet does not exist on dest host');
+
```
oetiker: to preserve present functionality, you would first check for autoCreation and then check for the datasets existance and then create it. since create will already throw an exception if it fails, there is no need todo anything anymore here.

oetiker:
I like the idea of an option to create missing datasets on the destination ...

lib/ZnapZend/ZFS.pm
Wouter0100: Added info message, since this shouldn't happen on a hourly basis. Only once per dataset, so yeah. :+1:

lib/ZnapZend/ZFS.pm
Wouter0100: Hmm, this will only throw an exception when the dataset isn't there, so this dataset is being skipped (else it will start the transfer, which throws a lof of errors). The createDataSet is only called when autoCreation is enabled and the dataset isn't there yet. 

I've added some extra information to the exception.

lib/ZnapZend/ZFS.pm
oetiker: to log debug messsages 

```perl
$self->zLog->debug("message ...");
```

Wouter0100: Hmm, do you want this to be debug instead of info? I've added the info below this reply.. :smile:

lib/ZnapZend/ZFS.pm
```diff
@@ -122,6 +123,25 @@ sub snapshotExists {
...
+    my $remote;
+
+    ($remote, $dataSet) = $splitHostDataSet->($dataSet);
+    my @ssh = $self->$buildRemote($remote,
+        [@{$self->priv}, qw(zfs create), $dataSet]);
+
+    print STDERR '# ' . join(' ', @ssh) . "\n" if $self->debug;
+
+    $self->zLog->info("create new dataset ($dataSet) on destination ($remote)");
+
```
oetiker: so you can remove the print STDERR

Wouter0100: Hmm, I thought in debug all commands are logged to STDERR? (or, it seems like)

oetiker: hihi thats what you get when I comment on code that I did not write myself ... ok in that case

Wouter0100: Awesome :smile:

bin/znapzend
```diff
@@ -13,7 +13,7 @@ my %featureMap = map { $_ => 1 } qw(oracleMode recvu pfexec sudo);
 
 sub main {
     GetOptions($opts, qw(help|h man debug|d noaction|n nodestroy features=s),
-        qw(daemonize pidfile=s logto=s loglevel=s runonce=s connectTimeout=s)) or exit 1;
+        qw(daemonize pidfile=s logto=s loglevel=s runonce=s connectTimeout=s autoCreation)) or exit 1;
```
oetiker: please update the documentation as well (see the pod section below)

Wouter0100: Oh, yeah. Good point. I'll also add docs to the end of the file with the new sub

Wouter0100: Added :+1:

lib/ZnapZend/ZFS.pm
```diff
@@ -289,6 +309,19 @@ sub sendRecvSnapshots {
...
+    #attemp a creation of the dataset when it doesn't exist
+    $dstDataSetExists = $self->createDataSet($dstDataSet) if $self->autoCreation && !$dstDataSetExists;
+
+    ($remote, $dstDataSet) = $splitHostDataSet->($dstDataSet);
+
+    #check if the dstDataSet exist on the destination (maybe after a creation)
+    !$dstDataSetExists
+        and Mojo::Exception->throw("ERROR: dataset ($dstDataSet) does not exist on "
+	    . "destination ($remote), use --autoCreation to "
```
hadfl: can you change this to something like:
```... 'does not exist' . ($remote ? " on destination ($remote)" :'') ...```

in case remote is undefined

Wouter0100: Good point, let me add that.

Wouter0100: Added!

oetiker:
thanks

PR #203, opened 2016-04-04
sendRecvSnapshots - don't cut dstDataSet in-place

sempervictus:
The current code in sendRecvSnapshots splits the dstDataSet var
early on in the execution chain to provide logging output and the
remote variable contents. This removes the remote information from
dstDataSet which is then passed to listSnapshots without the
remote data and the attempt to list the last snapshot occurs on the
local host, not over SSH on the remote target.

Address the issue by creating a dstDataSetPath variable to hold the
stripped target path while retaining the original dstDataSet with
the remote information attached for lastAndCommonSnapshots to use
in listSnapshots.

-----

Testing:
  4 hosts with waterfall replication topologies in debug mode


oetiker:
guess this is a regression from #194 ... thanks!

PR #205, opened 2016-04-05
Update README.md

tisc0:
Install perl-core for redhat like systems, to permit missing modules installation when make / make install


oetiker:
I integrated that info on a different line ... thanks!

PR #219, opened 2016-06-12
Add missing step to checkinstall README

lckarssen:
After cloning the Git repo, the clone is located in the `znapzend` directory, so the user should first `cd` into it before running the checkinstall script.


oetiker:
thanks

PR #221, opened 2016-06-20
Mail logger

leoj3n:
I probably missed some documentation changes; will double check later.

### DIFF
--- a/bin/znapzend
+++ b/bin/znapzend
@@ -13,7 +13,7 @@ my %featureMap = map { $_ => 1 } qw(oracleMode recvu pfexec sudo);
 
 sub main {
     GetOptions($opts, qw(help|h man debug|d noaction|n nodestroy features=s),
-        qw(daemonize pidfile=s logto=s loglevel=s runonce=s connectTimeout=s autoCreation)) or exit 1;
+        qw(daemonize pidfile=s logto=s@ loglevel=s runonce=s connectTimeout=s autoCreation)) or exit 1;
 
     #split all features into individual options
     if ($opts->{features}){
@@ -60,7 +60,7 @@ B<znapzend> [I<options>...]
  -d,--debug         print debug messages to STDERR
  -n,--noaction      run in simulation mode. does no changes to the filesystem
  --nodestroy        does all changes to the filesystem except destroy
- --logto=x          select where to log to (syslog::<facility> or <filepath>)
+ --logto=x          where to log to (syslog::<facility> or file::<path>, and/or pipe::<program>)
  --loglevel=x       define the log level when logging to file
  --pidfile=x        write a pid file when running in daemon mode
  --daemonize        fork into the background
@@ -92,16 +92,20 @@ configurations together with B<--debug>
 
 do all changes to the filesystem except destroy old snapshots
 
-=item B<--logto>={B<syslog::>I<facility>|I<filepath>}
+=item B<--logto>={B<syslog::>I<facility>|B<file::>I<path>|B<pipe::>I<program>}
 
 send logs out to either syslog or a logfile. Default is to send logs to
 B<syslog::daemon> when runing daemonized. When running in debug mode, the
-logs will go to STDERR by default.
+logs will go to STDERR by default. Logs can also be piped to a program of
+your choice, which will be done in batch every 60 seconds; you may specify
+multipe pipes that will be used in conjunction with either syslog or logfile.
 
 Examples:
 
- --logto=/var/log/znapzend.log
  --logto=syslog::daemon
+ --logto=file::/var/log/znapzend.log
+ --logto="pipe::mail -s 'ZnapZend Error' you@domain.com" --logto="pipe::mylinereader"
+ --logto=file::/var/log/znapzend.log --logto="pipe::mail -s 'ZnapZend Error' you@domain.com"
 
 =item B<--loglevel>={B<debug>|B<info>|B<warning>|B<err>|B<alert>}
 
--- a/lib/ZnapZend.pm
+++ b/lib/ZnapZend.pm
@@ -1,7 +1,7 @@
 package ZnapZend;
 
 use Mojo::Base -base;
-use Mojo::IOLoop::ForkCall;
+use Mojo::IOLoop::ReadWriteFork;
 use Mojo::Util qw(slurp);
 use Mojo::Log;
 use ZnapZend::Config;
@@ -391,15 +391,25 @@ my $sendWorker = sub {
 ### RM_COMM_4_TEST ###  return;
 
     #send/receive fork
-    my $fc = Mojo::IOLoop::ForkCall->new;
-    $fc->run(
+    my $rwf = Mojo::IOLoop::ReadWriteFork->new;
+    $rwf->run(
         #send/receive worker
         $sendRecvCleanup,
         #send/receive worker arguments
         [$self, $backupSet, $timeStamp],
         #send/receive worker callback
         sub {
-            my ($fc, $err) = @_;
+            my ($rwf, $err) = @_;
+
+            $self->zLog->on(
+                message => sub {
+                    my ($log, $level, @lines) = @_;
+                    for (@lines){
+                        print STDERR "$level $_\n";
+                    }
+                }
+            );
+            $self->zLog->warn('WARN FROM CHILD FORK!!');
 
             $self->zLog->warn('send/receive for ' . $backupSet->{src}
                 . ' failed: ' . $err) if $err;
@@ -412,9 +422,9 @@ my $sendWorker = sub {
     );
 
     #spawn event
-    $fc->on(
+    $rwf->on(
         spawn => sub {
-            my ($fc, $pid) = @_;
+            my ($rwf, $pid) = @_;
         
             $self->zLog->debug('send/receive worker for ' . $backupSet->{src}
                 . " spawned ($pid)");
@@ -422,10 +432,26 @@ my $sendWorker = sub {
         }
     );
 
+    # read child output
+    my $buffer = '';
+    $rwf->on(
+        read => sub {
+            my ($rwf, $chunk) = @_; # $buffer = both STDERR and STDOUT
+            $buffer .= $chunk;
+            while (my ($line) = $buffer =~ s/(.+)\n//){
+                for ($line){
+                   /error/i && do { $self->zLog->error($line); next; };
+                   /warn/i && do { $self->zLog->warn($line); next; };
+                   $self->zLog->notice($line);
+                }
+            }
+        }
+    );
+
     #error event
-    $fc->on(
+    $rwf->on(
         error => sub {
-            my ($fc, $err) = @_;
+            my ($rwf, $err) = @_;
 
             $self->zLog->warn($err) if !$self->terminate;
         }
@@ -443,16 +469,26 @@ my $snapWorker = sub {
 ### RM_COMM_4_TEST ###  return;
 
     #snapshot fork
-    my $fc = Mojo::IOLoop::ForkCall->new;
-    $fc->run(
+    my $rwf = Mojo::IOLoop::ReadWriteFork->new;
+    $rwf->run(
         #snapshot worker
         $createSnapshot,
         #snapshot worker arguments
         [$self, $backupSet, $timeStamp],
         #snapshot worker callback
         sub {
-            my ($fc, $err) = @_;
+            my ($rwf, $err) = @_;
             
+            $self->zLog->on(
+                message => sub {
+                    my ($log, $level, @lines) = @_;
+                    for (@lines){
+                        print STDERR "$level $_\n";
+                    }
+                }
+            );
+            $self->zLog->warn('WARN FROM CHILD FORK!!');
+
             $self->zLog->warn('taking snapshot on ' . $backupSet->{src}
                 . ' failed: ' . $err) if $err;
 
@@ -472,9 +508,9 @@ my $snapWorker = sub {
     );
 
     #spawn event
-    $fc->on(
+    $rwf->on(
         spawn => sub {
-            my ($fc, $pid) = @_;
+            my ($rwf, $pid) = @_;
         
             $self->zLog->debug('snapshot worker for ' . $backupSet->{src}
                 . " spawned ($pid)");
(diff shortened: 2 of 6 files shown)


bin/znapzend
```diff
@@ -107,6 +109,22 @@ Examples:
...
+ --mailto=admin@domain.com
+
+=item B<--mailprog>={I<filepath>}
+
+path to program to use for sending mail
+
+Examples:
+
+ --mailprog=/usr/bin/mail
+
```
oetiker: I guess you should mention that the code expects that the mail program accepts the `-s` argument

lib/ZnapZend.pm
```diff
@@ -87,13 +90,40 @@ has zLog => sub {
...
+            close $fh;
+        }
+            
+        $self->logBuffer([]);
+    };
+
+    if ($self->runonce){
+        #run immediately
+        $cb->();
+    }
```
oetiker: hmmm at that point in time there will be nothing to report I guess ...

lib/ZnapZend.pm
```diff
@@ -87,13 +90,40 @@ has zLog => sub {
...
+    else{
+        Mojo::IOLoop->timer(10 => $cb);
+    }
```
oetiker: at the moment you will send mail every 10 seconds ... I think it would be cool if you would send email if either (more than 1 minute has passed since the last email or no more new data has become available for more than 10 seconds) that should reduce the number of mails in case there is high logging activity

t/znapzend.t
```diff
@@ -53,6 +56,13 @@ sub runCommand {
     main();
 }
 
+sub mailLogContainsWARN {
+    open my $fh, "mail.log";
+    my @mail_log = <$fh>;
+    close $fh;
+    return (grep(/WARN/,@mail_log) ne 0);
+}
+
```
oetiker:
```perl
sub mailLogContainsWarn {
    open my $fh, '<',"mail.log";
    return ((grep /WARN/, <$fh>) != 0)
}
```

bin/znapzend
```diff
@@ -96,12 +96,16 @@ do all changes to the filesystem except destroy old snapshots
...
+logs will go to STDERR by default. Logs can also be piped to a program of
+your choice which will be done in a batch every 60 seconds; you may have
+multipe pipes and can be used with syslog or logfile by appending a comma.
 
 Examples:
 
- --logto=/var/log/znapzend.log
  --logto=syslog::daemon
+ --logto=/var/log/znapzend.log
+ --logto=pipe::mail -s 'ZnapZend Error' you@domain.com
```
oetiker: --logto="pipe::mail -s 'ZnapZend Error' you@domain.com"

lib/ZnapZend.pm
```diff
@@ -61,21 +63,48 @@ has zTime => sub { ZnapZend::Time->new() };
...
+        my $logto = $self->logto;
+        if (substr($logto, -1) eq '"' && substr($logto, 0, 1) eq '"' ||
+            substr($logto, -1) eq "'" && substr($logto, 0, 1) eq "'") {
+          $logto = substr $logto, 1, -1;
+        }
+
+        #log to user-defined logger(s)
+        my @logtos = split /,/, $logto;
+        foreach (@logtos){
+          /^(((?!::).)*)$/ && do {
```
oetiker: to stay 'in syntax' I would use
`append::`logfilename or `file::`logfilename

lib/ZnapZend.pm
```diff
@@ -61,21 +63,48 @@ has zTime => sub { ZnapZend::Time->new() };
...
+    my ($file, $syslog);
+
+    if ($self->logto){
+        #remove any wrapping quotes
+        my $logto = $self->logto;
+        if (substr($logto, -1) eq '"' && substr($logto, 0, 1) eq '"' ||
+            substr($logto, -1) eq "'" && substr($logto, 0, 1) eq "'") {
+          $logto = substr $logto, 1, -1;
+        }
+
```
oetiker: why would you want to remove quotes ? the shell will do that just fine ... such things would only be needed in dos :)

leoj3n: ah; I was testing using the test, that's why there was an issue; will remove this bit.

lib/ZnapZend.pm
leoj3n: I considered this, but wasn't sure about breaking the existing API.

bin/znapzend
```diff
@@ -96,12 +96,16 @@ do all changes to the filesystem except destroy old snapshots
...
+your choice which will be done in a batch every 60 seconds; you may have
+multipe pipes and can be used with syslog or logfile by appending a comma.
 
 Examples:
 
- --logto=/var/log/znapzend.log
  --logto=syslog::daemon
+ --logto=/var/log/znapzend.log
+ --logto="pipe::mail -s 'ZnapZend Error' you@domain.com"
+ --logto="/var/log/znapzend.log,pipe::mail -s 'ZnapZend Error' you@domain.com"
```
oetiker: .... `file::`

bin/znapzend
```diff
@@ -96,12 +96,16 @@ do all changes to the filesystem except destroy old snapshots
...
-logs will go to STDERR by default.
+logs will go to STDERR by default. Logs can also be piped to a program of
+your choice which will be done in a batch every 60 seconds; you may have
+multipe pipes and can be used with syslog or logfile by appending a comma.
 
 Examples:
 
- --logto=/var/log/znapzend.log
  --logto=syslog::daemon
+ --logto=/var/log/znapzend.log
```
oetiker: ... `file::`

lib/ZnapZend.pm
```diff
@@ -61,21 +63,41 @@ has zTime => sub { ZnapZend::Time->new() };
...
-    #check if we are logging to syslog
-    my ($syslog) = $self->logto =~ /^syslog::(\w+)$/;
-    # logging defaults to syslog::daemon (STDERR if runonce)
-    $syslog = 'daemon' if !$self->logto && !$self->runonce;
+    my ($file, $syslog);
+
+    if ($self->logto){
+        #log to user-defined logger(s)
+        my @logtos = split /,/, $self->logto;
+        foreach (@logtos){
```
oetiker: foreach -> for

lib/ZnapZend.pm
```diff
@@ -61,21 +63,41 @@ has zTime => sub { ZnapZend::Time->new() };
...
+          /^file::(.+)/ && do {
+              $file = $1;
```
oetiker: last;

lib/ZnapZend.pm
```diff
@@ -61,21 +63,41 @@ has zTime => sub { ZnapZend::Time->new() };
...
+          };
+
+          /^syslog::(\w+)$/ && do {
+              $syslog = $1;
```
oetiker: last;

lib/ZnapZend.pm
```diff
@@ -61,21 +63,41 @@ has zTime => sub { ZnapZend::Time->new() };
...
+          };
+
+          /^pipe::(.+)/ && do {
+              push @{$self->logPipes}, $1;
```
oetiker: last;

lib/ZnapZend.pm
```diff
@@ -61,21 +63,41 @@ has zTime => sub { ZnapZend::Time->new() };
...
+          };
```
oetiker: $file = $_;

leoj3n:
This is just about done. One thing you should double check before merging is that the `syslog::` option still functions as you intend it to.

leoj3n:
http://mojolicious.org/perldoc/perlipc#Safe-Pipe-Opens

oetiker:
yes ?

leoj3n:
@oetiker sorry, was just aggregating information on the current issue (of forked children not being able to communicate their status back to parent for batched logging to external pipes); essentially I was making a "read later" for myself or anyone else who might be listening.

If you have anything to add please do, because I don't see a clear path forward, as I'm not sure if the information in that (very interesting) linked article ultimately applies to forked children created like `Mojo::IOLoop::ForkCall->new`. If you can clarify what I should be looking at and trying (if you know) I would appreciate that.

oetiker:
@leoj3n oh sorry I thought it was clear ... 

what I suggest todo is to simple print any error found in the forked child to stderr, in a sensible format

```perl
    my $log = $self->zLog;
    $log->unsubscribe('message');
    $log->on(message => sub {
        my ($log, $level, @lines) = @_;
        for (@lines){
            print STDERR "$level $_\n";
        });
   });
```


the result of this will be that these errors, plus anything printed from ssh or from the zfs command at the remote end via ssh will end up in the parent, where read events will be generated.

```perl
my $buffer ='';
   $fc->on(read => sub {
        my($fork, $chunk) = @_; # $buffer = both STDERR and STDOUT
        $buffer .= $chunk;
        while ( my ($line) = $buffer =~ s/(.+)\n// ) {
            for ($line){
              /error/i &&  $self->zLog->error($line); next; };
              /warn/i &&  do { $self->zLog->warn($line); next; };
              $self->zLog->notice($line);
           }
        }
    });
    $fc->on(error => sub {
        my($fork, $error) = @_;
        $self->zLog->error("error $error");
    });
```

leoj3n:
Ah, should have guessed (or RTFM) that ForkCall would provide a read event. It's actually quite elegant of a solution, in its simplicity.

leoj3n:
@oetiker So, I assume I need to check if the current execution is a forked child, for the first bit of code that reassigns the message. How can I check if forked child so I can wrap that code with the check?

jberger:
@oetiker ForkCall does not natively provide a mechanism to report back to the parent while the child is still executing. This is by design since the closing of the pipe signals that the process is complete which solves a lot of timing problems. There are ways you can implement this communication on top of the ForkCall scaffolding, including opening another IO::Pipely (which is how ForkCall itself communicates) or else just opening a pipe (if you don't need say windows support). 

The other option would be to choose Mojo::IOLoop::ReadWriteFork which is a more full featured fork monitor. ForkCall is intended as a "fork, wait, collect results" module generally. If certainly can be made to do have a channel of communication back to the parent, but you have to implement that in your own code.

oetiker:
aargh sorry ... I meant Mojo::IOLoop::ReadWriteFork ...

leoj3n:
`perl znapzend.t`:

~~~
.
..
...
# zfs list -H -o name -t filesystem,volume backup/destfail
[Sun Jun 26 12:25:43 2016] [warn] destination 'backup/destfail' does not exist or is offline. will be rechecked every run...
[Sun Jun 26 12:25:43 2016] [info] found a valid backup plan for tank/source...
[Sun Jun 26 12:25:43 2016] [warn] WARN FROM CHILD FORK!!
warn WARN FROM CHILD FORK!!
[Sun Jun 26 12:25:43 2016] [debug] snapshot worker for tank/source spawned (ZnapZend=HASH(0x28313e0)->pid)
debug snapshot worker for tank/source spawned (ZnapZend=HASH(0x28313e0)->pid)
Can't locate object method "pid" via package "ZnapZend" at /root/zz/znapzend/t/../lib/ZnapZend.pm line 376.
# Tests were run but no plan was declared and done_testing() was not seen.
# Looks like your test exited with 255 just after 3.
~~~

https://github.com/oetiker/znapzend/pull/221/commits/7c3f931fc51dd8c86dead15f157f2a482125c3cb#diff-10b1b71b6b019e1f1c045d84a69aadd9R376

lib/ZnapZend.pm
```diff
@@ -391,18 +476,50 @@ my $snapWorker = sub {
...
+
+            $self->zLog->warn($err) if !$self->terminate;
+        }
+    );
+
+    $rwf->on(
+        close => sub {
+            my($rwf, $exit_value, $signal) = @_;
+
+            $self->log->error("dbSlave quit with $exit_value and signal $signal");
```
oetiker: this should only be an error if there is an exit_value != 0 or a $signal != 0

lib/ZnapZend.pm
```diff
@@ -391,18 +476,50 @@ my $snapWorker = sub {
...
+    $backupSet->{snap_pid} = $rwf->pid;
+
+    # read child output
+    my $buffer = '';
+    $rwf->on(
+        read => sub {
+            my ($rwf, $chunk) = @_; # $buffer = both STDERR and STDOUT
+            $buffer .= $chunk;
+            while (my ($line) = $buffer =~ s/(.+)\n//){
+                for ($line){
```
oetiker: maybe prepend the process id

`$line = "[".$rwf->pid."] ".$line;`

lib/ZnapZend.pm
```diff
@@ -391,18 +476,50 @@ my $snapWorker = sub {
...
+        . " spawned (" . $rwf->pid . ")");
+    $backupSet->{snap_pid} = $rwf->pid;
+
+    # read child output
+    my $buffer = '';
+    $rwf->on(
+        read => sub {
+            my ($rwf, $chunk) = @_; # $buffer = both STDERR and STDOUT
+            $buffer .= $chunk;
+            while (my ($line) = $buffer =~ s/(.+)\n//){
```
oetiker: maybe prepend the process id

`$line = "[".$rwf->pid."] ".$line;`

leoj3n:
OK, for anyone else who may be following along...

@oetiker figured out that we needed to wait for the IO loop before querying the PID of the child fork.

Child warns/errors are currently being redirected to STDERR, which is a step in the right direction, but the parent ZnapZend process is not yet successfully reading in that STDERR output.

leoj3n:
It seems none of the event handlers are being called (`read`/`error`/`close`).

I have tried changing `$rwf->start()` to `$rwf = $rwf->start()` as well as changing:
```
    $rwf->start(
        program_args => [$self, $backupSet, $timeStamp],
        program => $sendRecvCleanup,
        conduit => 'pipe',
    );
```
to remove `conduit => 'pipe'` like:
```
    $rwf->start(
        program_args => [$self, $backupSet, $timeStamp],
        program => $sendRecvCleanup,
    );
```
But still no event handlers are called.

lib/ZnapZend.pm
```diff
@@ -391,18 +480,55 @@ my $snapWorker = sub {
...
+
+            $self->zLog->warn($err) if !$self->terminate;
+        }
+    );
+
+    $rwf->on(
+        close => sub {
+            my($rwf, $exit_value, $signal) = @_;
+
+            $self->log->error("child fork quit with $exit_value and signal $signal")
```
oetiker: log -> zLog ?

jimklimov:
Saw this PR after posting my suggestion #499 but after re-reading this, not sure it would be a duplicate :) PR #221 seems to add ability to log *everything* emitted into a file or pipe, same as done for console (stderr/stdout) or syslog, and #499 concerns about alerting to fail to send and so refusing to clean up origin (which may lead to it getting too full).

PR #226, opened 2016-07-27
add possibility to specify delay before sending snaps

renarded:


lib/ZnapZend.pm
```diff
@@ -168,6 +168,12 @@ my $sendRecvCleanup = sub {
     #no HUP handler in child
     $SIG{HUP} = 'IGNORE';
 
+    if ($backupSet->{zend_delay} and $backupSet->{zend_delay} > 0){
+        $self->zLog->info("waiting $backupSet->{zend_delay} seconds before sending snaps...");
+        sleep $backupSet->{zend_delay};
+        $self->zLog->info("resume sending action");
+    }
```
oetiker: you may want to add $backupSet->{src} to the log lines, so that it is clear about which backupset you are talking ...

doc/znapzendzetup.pod
```diff
@@ -13,6 +13,7 @@ where 'command' is one of the following:
             [--pre-snap-command=<command>] \
             [--post-snap-command=<command>] \
             [--tsformat=<format>] --donotask \
+            [--send-delay=<time>] \
             SRC plan dataset \
```
oetiker: this is a generated file ... you don't need to patch it ...

oetiker:
thanks for your patch!

PR #227, opened 2016-07-28
disable StrictHostKeyChecking on ssh connexion

Houlbek:
The aim iz to make industrialization eazier.


oetiker:
in recent months we have actually removed all non essential options from our ssh calls, as peoples requirements are so diverse... the idea is to use your `.ssh/config` file to customize the ssh settings

PR #230, opened 2016-08-10
add option to avoid sending unwanted snapshots (using send -i and -I)

renarded:
Here is an example for the use of --strict-send option :

Suppose i have those snapshots on source server (some created by znapzend, some not) :
tank/home@2016-08-09_00-00-00
tank/home@nightly.20160809
tank/home@homemadesnap
tank/home@2016-08-09_03-00-00
tank/home@uselesssnapshotforthisexample
tank/home@2016-08-09_06-00-00
tank/home@2016-08-09_09-00-00
tank/home@2016-08-09_12-00-00
tank/home@homemadesnap2

And option --tsformat is %Y-%m-%d_%H-%M-%S.

So, I want to send snapshots with znapzend.
Suppose the last common snapshot with backup server is tank/home@2016-08-09_00-00-00.

Then, " send -I tank/home@2016-08-09_00-00-00 tank/home@2016-08-09_12-00-00 " will be used. Problem is : option -I will send all snapshots between those two snaps, included the ones i did not want (for this example i only want to receive znapzend snaps)

With use of --strict-send option, multiple send commands will be executed if needed : 
send -i tank/home@2016-08-09_00-00-00 tank/home@2016-08-09_03-00-00
send -i tank/home@2016-08-09_03-00-00 tank/home@2016-08-09_06-00-00
send -I tank/home@2016-08-09_06-00-00 tank/home@2016-08-09_12-00-00

So I only have snapshots "strictly" matching tsformat on backup server.

### DIFF
--- a/bin/znapzendzetup
+++ b/bin/znapzendzetup
@@ -112,7 +112,7 @@ sub main {
 
     /^create$/ && do {
         GetOptions($opts, (qw(recursive|r donotask pfexec sudo mbuffer=s mbuffersize=s),
-            qw(tsformat=s pre-snap-command=s post-snap-command=s send-delay=i))) or exit 1;
+            qw(tsformat=s pre-snap-command=s post-snap-command=s send-delay=i strict-send))) or exit 1;
 
         $cfg{enabled}       = 'on';
         $cfg{recursive}     = $opts->{recursive}            ? 'on' : 'off';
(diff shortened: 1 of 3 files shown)


dylanPowers:
I'd argue that this should be the default behavior simply because if a snapshot is not specified in a remote plan, that snapshot shouldn't end up on the remote server.
@renarded you still around to resolve the merge conflict? I kind of want to play around with this patch in my own fork

jimklimov:
The idea and premise looks reasonable, at least for cases described here. In my own usage patterns I make manual snapshots on purpose (and/or as part of e.g. system updates) so usually benefit from having the `-I` as default and know all my history and its turning points are replicated, but I see how this can be a problem for other use-cases (e.g. volumptious snapshots from stages of some experimental data re-arrangement and/or dataset cloning), so +1 for the idea with the filter (and hope it gets updated vs. conflicts and merged eventually).

* pun intended

jimklimov:
The PR linked above does not *fully* cover the premise of this PR (e.g. the filtering and construction of a list of snapshots to send in order is still an interesting concept), but the basic `-i/-I` opt-in is covered.

PR #233, opened 2016-09-24
Added destination pre and post send/receive commands.

dunron:
To be used for waking up destination devices or putting remote backup disks on- and offline again. Or send geli-keys from the local server to the remote host for very secure backups.


bin/znapzendzetup
```diff
@@ -637,6 +651,7 @@ S<Dominik Hassler E<lt>hadfl@cpan.orgE<gt>>
 
 =head1 HISTORY
 
+2016-09-23 had Destination pre and post send/receive commands
```
oetiker: had is the sign of @hadfl  ... please use your own :)

dunron: I *had* no idea.. :-J

oetiker:
thanks!

PR #243, opened 2016-11-21
updated checkinstall script and docs

morph027:
Automated the checkinstall script, added the already existing method to main README.


morph027:
Contributing to https://github.com/Gregy/znapzend-debian ... the checkinstall thing is not as reliable as it pretends to be ...

PR #245, opened 2016-12-01
Changed SMF dependency to multi-user

dglushenok:
znapzend starts too early in case of ZFS over iSCSI with delayed zpool import.


oetiker:
sounds good ...

PR #251, opened 2017-01-05
per destination pre/post-send-command improvements

dunron:
Added an environment variable that can be used by the scripts. Within the script it is now possible to determine by which worker it was called. Also added some documentation and made it overall more consistent with the existing pre/post-snap-command feature.


hadfl:
looks good to me. what do you think @oetiker ?

oetiker:
thanks

dunron:
thank you!

PR #252, opened 2017-01-06
cleanup gitter link from referral url source param

JMoVS:
The URL previously contained referrer parameters. I cleaned it up so it doesn’t contain that anymore.

Signed-off-by: Justin Scholz <git@justinscholz.de>


oetiker:
thanks

Issue #253, opened 2017-01-12
Daemon refuses to start if there are no backup plans


Gregy:
Hi,
I run across this while trying to integrate znapzend into a bigger system. The daemon refuses to start if there are no backup plans defined.
```
ERROR: no backup set defined or enabled, yet. run 'znapzendzetup' to setup znapzend
```
Wouldn't it be better to issue a warning but start znapzend even if there are no plans? Later I could add a plan and just send a signal to znapzend to reload configuration.

Right now if I add a new plan I have to check if znapzend is running in which case I have to send a signal. And if znapzend is not running I have to start it.

Thank you

oetiker:
hmmm yes, I could understand if the thing would keep running in the background, after complaining about the missing plans ... PR welcome.

PR #256, opened 2017-01-16
Makes daemon continue even if there are no enabled backup sets

Gregy:
Hi,
this PR relates to #253 . It makes the daemon continue running even if there are no backups defined.

I have never touched perl code before but this change seems really simple. All tests are passing and manual testing shows desired behavior.

Thank you


hadfl:
thank you!

PR #280, opened 2017-07-17
Added compression feature to add -Lce options to zfs send.

hessenpepper:
My first contribution to OSS.

-L and -e were added to zfs around the same time as -c, so I've put them all together.

### DIFF
--- a/bin/znapzend
+++ b/bin/znapzend
@@ -9,7 +9,7 @@ use Pod::Usage;
 use ZnapZend;
 my $VERSION = '0.dev'; #VERSION
 my $opts = {};
-my %featureMap = map { $_ => 1 } qw(oracleMode recvu pfexec sudo);
+my %featureMap = map { $_ => 1 } qw(oracleMode recvu pfexec sudo compression);
 
 sub main {
     GetOptions($opts, qw(help|h man debug|d noaction|n nodestroy features=s),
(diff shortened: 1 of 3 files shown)


bin/znapzend
```diff
@@ -164,6 +164,15 @@ use 'pfexec' for zfs commands (e.g. when running znapzend as non-priviledged use
...
 use 'sudo' for zfs commands (e.g. when running znapzend as non-priviledged user)
 
+=item compression
+
+use 'compression' to add options -Lce to the zfs send command
+If a source and destination volume are both using compression, zfs send will, by
+default, uncompress the data before sending. Zfs recv will then compress it again before 
+writing it to disk. Using -c will skip the unnecessary uncompress-compress stages.
+This decreases CPU load on both source and destination as well as reduce network
+bandwidth. The -L and -e options were thrown in for good measure.
```
oetiker: maybe mention a key word for L and e in brackets ... for good measure is a bit very flippant I think

oetiker:
thanks!

hadfl:
if you'd change https://github.com/oetiker/znapzend/pull/280/files#diff-33cf5e498f84195d4308abccb7c59d70R314

to an array `my @sendOpt = $self->compressed ? qw(-Lce) : ();`

you would not need the checks `if($sendOpt){` but could just do:
`@cmd = ([@{$self->priv}, 'zfs', 'send', @sendOpt, ...`

as the empty array (in case compressed option is not set) won't introduce a member to the command array.

bin/znapzend
hessenpepper: I made that change and fixed a problem with a little help from @hadfl

oetiker:
thanks for the new feature!

PR #283, opened 2017-08-18
Fix Mojo::Util::slurp is DEPRECATED

drscream:
The slurp function has been removed in the latest Mojo release for that reason it's required to use the path function to slurp.


mas90:
Mojo::File doesn't exist in the version of Mojolicious that znapzend currently uses.  I've opened an alternative PR #395 that doesn't use any Mojo slurp.

oetiker:
merged #395

PR #294, opened 2017-10-16
Allow runonce to run across all datasets

dylanPowers:
The motivation behind this was to run znapzend as a cron job.


oetiker:
thanks

PR #299, opened 2017-10-22
Add `resume` feature for resumable transfer

aarononeal:
Adds `-s` to the receive command to support a resumable transfer. Future work could extend to support resume token on send.


oetiker:
are you planning on completing this feature ? automatic transfer resumption especially for initial sync might be something many people would value I guess

aarononeal:
I think what’s here is valuable even without the auto resume part since resume can be initiated manually. I used it, for exmaple, to do a runonce on my initial sends and resume them manually when they failed. So it’s a good first step and I think worth integrating as is.

With this in place I was hoping someone more versed in Perl and the codebase might add the other part in the future or advise on the best way to programmatically retrieve the token property from the filesystem. That said, if there’s no one else, I might get to this over the next couple weeks since it should be as easy as getting that property and using it to change the send command.

oetiker:
I will be glad to advise if you run into trouble ...

Harvie:
I have to agree. It makes sense to do `zfs recv -s` every time. Even when there's no way to automaticaly resume after crash. If you don't use `-s` there's no chance to resume it manualy. If you use `-s` user can at least try to resume it manualy.

oetiker:
anyone sees a downside to running send with -s generally (maybe test if supported I guess)

Harvie:
bump

oetiker (approved):

PR #314, opened 2018-01-12
`zfs send -i` vs. `zfs send -I`

znek:
`zfs send -I` in the context of znapzend is plain wrong, it really needs to be `zfs send -i` (also note the discussion at https://github.com/oetiker/znapzend/pull/230).

Using `-I`, totally unrelated snapshots are being sent to dst (if there are any) and even worse, such snapshots will interfere with the expiry (snapshot cleanup)  mechanism, which is the reason why I stumbled across this problem.


lib/ZnapZend/ZFS.pm
```diff
@@ -335,7 +335,7 @@ sub sendRecvSnapshots {
 
     my @cmd;
     if ($lastCommon){
-        @cmd = ([@{$self->priv}, 'zfs', 'send', @sendOpt, '-I', $lastCommon, $lastSnapshot]);
+        @cmd = ([@{$self->priv}, 'zfs', 'send', @sendOpt, '-i', $lastCommon, $lastSnapshot]);
```
oetiker: if znapzend has been creating local snapshots for some time, without being able to send them to the destination (dest was down maybe) then, calling -I will ensure that all the snapshots that did NOT go out to dest will be sent ... 

It is true, that if you create your own snapshots by hand, they will go out too ... so if you want to make sure that only znapzend made znapshots stay on dest, you would have to implement an option to clean unwanted snapshots from the destination ...

znek: Ok, which leads us back to https://github.com/oetiker/znapzend/pull/230. I do see your point, which suggests that an option like the one (or similar to the one) proposed in https://github.com/oetiker/znapzend/pull/230 should be added. I remain convinced, however, that the default should be `-i`. In my case, `-I` has proven to be a problem and indeed required me to have a periodic script that destroyed unwanted snapshots. I fail to see why that should be a default (and expected) behavior.

oetiker: the default is -I because in regards to backup it is better to err on the side of caution ... as I said, I think the simplest solution would be to have the option of removing any snapshots from destination that do not comply with the naming scheme chosen in your znapzend setup.

jimklimov:
I believe PRs #459 and #466 have fulfilled the goals of this request in a more flexible and backwards-compatible manner.

@znek: Can you verify if the current znapzend master branch can do what you want like `znapzend --features=skipIntermediates --runonce=tank/source`, and if yes - close this PR?

znek:
Thanks for the heads up! I'm currently lacking the time to fully investigate, but by just reading #459 and #466 I fully agree. Closing this for the sake of generating less noise.

Issue #297, opened 2017-10-17
Missed backups for daylight savings or TZ changes


dylanPowers:
When timezone related backward jumps in time occur, znapzend will try and fail to overwrite old snapshots. Meanwhile, no new snapshots will be made. I think by default, the timezone's offset should be appended to the string:
```
%z | ISO 8601 offset from UTC in timezone (1 minute=1, 1 hour=100)If timezone cannot be determined, no characters | +100
```
Or just default to UTC?

oetiker:
you can configure the string used by znapzend ...

oetiker:
:) PR is welcome ... note though not all systems may support %z

jimklimov:
Code is merged, should the issue be closed?

PR #315, opened 2018-02-05
Add support for %z specifier in tsformat option

atj:
This implements the functionality requested in issue #297.

If "%z" isn't supported then checkTimeFormat should call die(), however I don't have access to such a platform to confirm this.


atj:
@hadfl: any chance you can review this PR? Thanks!

PR #316, opened 2018-02-06
Allow skipping of snap create/send on pre-command failure

atj:
This commit implements two new boolean command line options:

* --skipOnPreSnapCmdFail
* --skipOnPreSendCmdFail

These options cause znapzend to skip snapshot creation or replication if
the respective pre-command returns a non-zero exit code or is killed by
a signal.


atj:
I'm receptive to comments and/or suggestions on this change. Particularly if someone can suggest some better option names!

oetiker:
I like it, @hadfl what do you think ?

hadfl:
lgtm

GabrieleV:
For me works correctly.

PR #321, opened 2018-02-19
Some more light edits

grantwwu:
About the "startup sequence" bit - you mean to start it on startup as a daemon, like any other system service, right?


oetiker:
thanks

PR #348, opened 2018-05-05
Further fix for service integrations

jimklimov:
Recent use of the `@exec_prefix@` expansion caused a `${prefix}` or even `NONE` to appear in the produced files. Pick up the double-eval trick from other projects to provide a static final string value for substitutions.


oetiker:
thanks

rcproam:
Unfortunately this is still an issue with v0.19.1.

./configure --prefix=/opt/znapzend

...yields a /init/znapzend.service file with ExecStart=NONE/bin/znapzend $ZNAPZENDOPTIONS on Debian Stretch

jimklimov:
Confirm that on a clean system, PR pending. Thanks for the report. Guess having it work in development was a peculiarity of workstation setup or the args I passed...

rcproam:
Cool thanks for the prompt response! :-)

Issue #347, opened 2018-05-05
feature request: Detailed documentation


mehanzhi:
Hello,
 
1. I spend almost 2 days trying to figure what this quote from the documentation. 
> A plan is required both for the source and the destination datasets.

I have created a plan ONLY on my running system and put there as a destination my backup server. 
It still worked. It didn't just worked but in fact it transferred all the snapshots on this dataset even the ones not created by znapzend. 
Actually I haven't found information at all about what is source and what is destination in the context of znapzend. Now it appears that I may have a plan for my destination but on the source it may be a different plan and I can still see src_plan and dst_plan both on my source system!?!?! And when I login on my destination server I see neither source or destination plan but somehow I have the snapshots transferred there. 

2. The plan I created involved ONLY daily and weekly backups and no frequent backups but I still see znapzend doing snapshots every hour. None of that was described in the documentation. 
I spend 3 hours searching for the answer of the question "if I set a daily backup at what time the snapshots will be created and sent to the destination???" Yes - it will be once at day but will it be 3 o'clock at the night or it will be right in the middle of the business hours? Didn't see anything in the documentation and no configuration switch for this purpose. 
The plan is: 
zroot  org.znapzend:dst_a_plan     30days=>24hours,90days=>1weeks,10years=>1months  local
But I still see snapshots being created every hour. 

3. There are stuff about mbuffer in the documentation and in the man page but no further description at all - do we need them? how and why are they used? How much memory should we use for mbuffers? do we need a formula for calculation? what is the impact if we do not use, or do not specify mbuffers? 
I have used backups zfs snapshots but using manual shell scripting. And I never had to use mbuffers for this. 
There was a question posted under issues with request for further information about mbuffers but there's no answer at all. 

I am sure the behavior I have seen is quite normal and probably all according to the purpose of znapzend. But !!!please!!! write some good documentation that will make clear how to use this product. These few examples are good but there are a lot of questions every person would ask and there are no answers at all. 

Thank you

jimklimov:
Thanks for your comment, I hopefully answered the mbuffer question while documenting my woes ;)

Issue #335, opened 2018-04-19
Do I have to use mbuffer?


gusioo:
Hi,

I have a question about mbuffer in znapzend: could anyone tells me when I must use mbuffer and what determines the size of mbuffer? Also what are the benefits and disadvantages of using mbuffer?

Best regards.

jimklimov:
Thanks for your comment, I hopefully answered the mbuffer question while documenting my woes ;)

gusioo:
Hi,

thanks, now it's clear.

Best regards.

Harvie:
How bad it will get if i don't use mbuffer? My replication seems to be very slow, i don't even think that some buffer can help so much... I guess i will try enabling it...

PR #350, opened 2018-05-05
Assorted Readme/znapzendzetup docs improvements

jimklimov:
Revised after some troubles I had during my first ever setup - hopefully helpful to other newcomers.

### DIFF
--- a/README.md
+++ b/README.md
@@ -92,7 +92,7 @@ znapzend --daemonize
 ```
 
 Best practice is to integrate znapzend into your system startup sequence, but you can also
-run it by hand. See the init/README.md for some inspiration.
+run it by hand. See the [init/README.md](init/README.md) for some inspiration.
 
 Statistics
 ----------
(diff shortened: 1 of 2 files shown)


jimklimov:
This PR should also address parts of #335 and #347 issues.

One thought I had while writing this (but did not have time to check in sources): the receiving side might be lagging, as well as the sending side, while `zfs send` or `zfs recv` is spinning on some lock or very busy doing its work. When I scripted large transfers over network (ssh, netcat, etc.) I had mbuffer running on both sides - so even if both zfs'es were slow, the two buffers still did not waste time transferring data over the network.

Does znapzend support such setup? Should there be options for it (path and config for mbuffer on DST #X)? Does the ability to configure mbuffer+port somehow cover this, with or without SSH (docs are still not clear, and who/how starts the networked daemon(s) for mbuffer)?

oetiker:
thanks

PR #352, opened 2018-05-08
znapzend.xml.in : do not go into maintenance if "zfs" coredumps

jimklimov:
Assuming that the `znapzend` (script) calls a system program in a loop, and that program returns well or dies, the script should handle it somehow, e.g. go to the next iteration of the loop, or abort with error and cause service restart. Either way, for the hands-off end-user, the clock goes on ticking and snaps taken.


oetiker:
thanks

PR #354, opened 2018-05-09
Allow use of /etc/default/znapzend for extra znapzend parameters in init files

asciiphil:
This brings the init files into alignment with standard packaging practices for various distributions.  It allows the init files to be included in a generic package while giving package users the latitude to pass their own parameters to the znapzend daemon.


oetiker:
thanks

Issue #356, opened 2018-05-09
on macOS znapzend does not startup correct at system startup


rwilkey:
Because of how Open ZFS on OSX mounts pools, znapzend starts up before any pools are mounted. This condition causes znapzend to go inactive and will only become active after a HUP signal is sent. This defeats the purpose of running with startup.  Also, launchd (macOS startup or init replacement) does not have a mechanism for resolving service dependencies so I haven't figured out how to resolve this).

I did find a solution by patching ZnapZend.pm to die on initialization if there is nothing to do (which is the case if no pools are mounted).  Then I can have launchd restart znapzend at some time interval, until the pools become available.  This was tested by exporting all the pools then starting up znapzend to observe the effects.

So should I make a pull request and show you my work?

Issue #322, opened 2018-02-20
non-replicated snapshots deleted if destination is down


healthian:
Znapzend will delete snapshots on SRC that have not been replicated to DST if DST is down. This occurs even to the point of deleting the last common snapshot between SRC and DST, forcing future replications to start over with a full replication.

Example:
SRC snapshots: A, B, C, D
DST snapshots: A, B

If DST is offline for too long, SRC snapshot A will be deleted, followed by B, then C, etc, while new snapshots E, F, G, etc will be created. However, once B is deleted, no more incremental replications will be able to take place, and a full resend of the entire SRC will be required to reestablish the DST backup.

The error appears to be in ZnapZend.pm, in the sendRecvCleanup method. When looping through all destinations, some error cases call next to skip the remainder of the loop without setting the $sendFailed flag, and before any send/recv can be executed (where the $sendFailed flag is set).

healthian:
I think this would fix it, but more people should test it.

```
diff --git a/lib/ZnapZend.pm b/lib/ZnapZend.pm
index 235bbd2..76b35e0 100644
--- a/lib/ZnapZend.pm
+++ b/lib/ZnapZend.pm
@@ -248,6 +248,7 @@ my $sendRecvCleanup = sub {
                 $self->zLog->warn("command \'" . $backupSet->{"dst_$key" . '_precmd'} . "\' failed");
                 if ($self->skipOnPreSendCmdFail){
                     $self->zLog->warn("skipping " . $backupSet->{"dst_$key"} . "due to pre-command failure");
+                    $sendFailed = 1;
                     next;
                 }
             }
@@ -273,6 +274,7 @@ my $sendRecvCleanup = sub {
                 $backupSet->{"dst_$key" . '_valid'} or
                     $self->zLog->warn("destination '" . $backupSet->{"dst_$key"}
                         . "' does not exist or is offline. ignoring it for this round...");
+                $sendFailed = 1;
                 next;
             };
         }
@@ -300,7 +302,11 @@ my $sendRecvCleanup = sub {
                     }
                 }
             }
-        } 
+        }
+
+        # do not destroy data sets on the destination, or run post-send-command, unless all operations have been successful
+        next if ($sendFailed)
+
         for my $srcDataSet (@$srcSubDataSets){
             my $dstDataSet = $srcDataSet;
             $dstDataSet =~ s/^\Q$backupSet->{src}\E/$backupSet->{$dst}/;

```

healthian:
[...]

PR #358, opened 2018-05-10
changes for macOS Startup

rwilkey:
These changes are proposed as solution to oetiker/znapzend issue #356 and maybe issue #322.

1. Change znapzend startup behavior to die if no zfs snap or backup sets are found.
2. Update init directory to add a launchd template and modify README.md to explain use.
3. Update main README.md to mention XCode command-line tools requirement.


rwilkey:
Ok. From what I see in the CI results. it looks like adding the org.oetiker.znapzend.plist.in file for launchd caused problems.  I did not modify any config or makefiles when I built this locally, as this is beyond my capability and probably authority!

lib/ZnapZend.pm
```diff
@@ -129,7 +129,7 @@ my $refreshBackupPlans = sub {
     $self->backupSets($self->zConfig->getBackupSetEnabled($dataSet));
 
     @{$self->backupSets}
-        or $self->zLog->warn("No backup set defined or enabled, yet. run 'znapzendzetup' to setup znapzend");
+        or die "No backup set defined or enabled, yet. run 'znapzendzetup' to setup znapzend";
```
oetiker: can you add a .. \n" at the end of your string so that the error message looks nice in the output.

oetiker:
@rwilkey your launchd.plist.in was fine ... no need to remove it ... I like it much better than large code blobs in the readme ... the reason for the test failure is that you are introducing a new error condition which used to be a warning condition ... you have to update the tests accordingly ... see https://travis-ci.org/oetiker/znapzend/jobs/377530186#L1232

lib/ZnapZend.pm
rwilkey: I think I did.

oetiker:
the entry in configure.ac is now missing from your patch ...

also you can call the file org.znapzend since we own this domain.

rwilkey:
@oetiker I see your comments. I assume from most recent, that you want the plist file to be named 'org.znapzend.plist.in' ?
also I see the test error in the log, but am unsure how to correct. I tried changing the test to look for 255 return but that did not do it (or the change never made it in).
I will look at the autoconf.ac

it appears merging is blocked so how do I proceed?

oetiker:
well for one you have to add the plist.in file to configure.ac so that a plist (without the .in) gets generated.

rwilkey:
@oetiker looks like this may be cleaned up. tests are passing and expecting the 'die' condition. Please review when you have time.

Thanks

oetiker:
thanks

rwilkey:
thanks for all the help in bearing with me and getting this done.

PR #372, opened 2018-07-16
Note about ZFS permissions for a non-priv user

flixman:


oetiker:
hmmm I would assume that the best place for such a note would be https://github.com/oetiker/znapzend/blob/master/bin/znapzendzetup

PR #374, opened 2018-08-14
Minor fix to supress an error message when doing 'znapzendzetup list'…

ShaunMaher:
… if a property is greater than 16 characters long

Credit to Paul Pettigrew of Mach Technology.


oetiker:
thanks

Issue #154, opened 2015-06-10
znapzendzetup disable on child


odoucet:
I've created a znapzend policy on 'mypoolname/user1' with recursive=ON.
I have several children on it : 

```
mypoolname/user1/webdata
mypoolname/user1/sqldata
mypoolname/user1/olddata
```

I would like to disable (temporarily or not) znapzend on a specific child, let's say "olddata". This cannot be done : 

```
znapzendzetup disable mypoolname/user1/olddata
ERROR: cannot disable backup config for mypoolname/user1/olddata. Did you create it?
```

If I try to edit the property myself it does not work : 

```
zfs set org.znapzend:enabled=off mypoolname/user1/olddata
```

And now "disable" shows an other message : 

```
znapzendzetup disable mypoolname/user1/olddata
ERROR: property recursive not set on backup for mypoolname/user1/olddata
```

hadfl:
by now, znapzend does not support to exclude child datasets when running w/ a recursive policy. you'll have to set up all child datasets individually (non recursive) if you want to exclude some of them....

odoucet:
this can be labeled as "feature-request" so, because I see znapzend is already looping on each child datasets for zfs send/recv. It may be simple to skip one if  org.znapzend:enabled=off
I'll dig into the source code and see if I can make a pull request on this.

jjlawren:
@odoucet, did you get anywhere with this?

I think the request would be interesting even if it only applied to overriding the DST option. I can live with local snapshots for transient datasets as long as they're not sent to a backup server.

odoucet:
did not make any progress at all... I was busy with something else .. Then totally forgot about this one.

morph027:
Right now, i'm keeping 2 datasets  for this.

e.g.

```
tank/critical
tank/non-critical
```

```tank/critical``` is being znapped recursively, the other won't get znapped anyway.

```
zfs rename -p tank/critical/disable-me tank/non-critical/disable-me
```

johnramsden:
This would be a killer feature.

flixman:
I am not good in perl, but I understand that the way to go is to do the regular snapshot and, right after that (if recursive), check which child ZFS have the flag, e.g., org.znapzend:enabled set to off.... nope?

oetiker:
yes, the send operations are 'per fs' anyway, so I think the change should not be all the big

flixman:
[...]

PR #379, opened 2018-09-23
Addressing issue #154

flixman:
As the sending of the ZFS streams is per stream, we can disable children ZFS from being snapshot/sent by setting the property "org.znapzend:enabled" to "off" on them. For the dataset being processed, in case the recursive flag is set, its children will be checked and their recently created snapshots removed.

### DIFF
--- a/lib/ZnapZend.pm
+++ b/lib/ZnapZend.pm
@@ -435,7 +435,7 @@ my $createSnapshot = sub {
 
         # restrict the list to the datasets that are descendant from the current
         my @dataSetList = grep /^$backupSet->{src}($|\/)/, @{$self->zZfs->listDataSets()};
-        if ( scalar @dataSetList > 0 ) {
+        if (@dataSetList) {
 
             # for each dataset: if the property "enabled" is set to "off", set the
             # newly created snapshot for removal
@@ -450,13 +450,14 @@ my $createSnapshot = sub {
                 # if the property does not exist, the command will just return. In this case,
                 # the value is implicit "on"
                 $prop = <$prop> || "on";
-                if ( substr($prop, 0, length('off')) eq 'off' ) {
+                chomp($prop);
+                if ( $prop eq 'off' ) {
                     push(@dataSetsExplicitelyDisabled, $dataSet . '@' . $snapshotSuffix);
                 }
             }
 
             # remove the snapshots previously marked
-           if ( scalar @dataSetsExplicitelyDisabled > 0 ){
+           if ( @dataSetsExplicitelyDisabled ){
                $self->zLog->info("Requesting removal of marked datasets: ". join( ", ", @dataSetsExplicitelyDisabled));
                $self->zZfs->destroySnapshots(@dataSetsExplicitelyDisabled);
            }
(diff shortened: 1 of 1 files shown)


lib/ZnapZend.pm
```diff
@@ -427,6 +427,42 @@ my $createSnapshot = sub {
...
+                
+                # get the value for org.znapzend property
+                my @cmd = (@{$self->zZfs->priv}, qw(zfs get -H -s local -o value org.znapzend:enabled), $dataSet);
+                print STDERR '# ' . join(' ', @cmd) . "\n" if $self->debug;
+                open my $prop, '-|', @cmd;
+
+                # if the property does not exist, the command will just return. In this case,
+                # the value is implicit "on"
+                $prop = <$prop> || "on";
+                if ( substr($prop, 0, length('off')) eq 'off' ) {
```
oetiker: `if ( $prop eq 'off')` should be fine :)

lib/ZnapZend.pm
```diff
@@ -427,6 +427,42 @@ my $createSnapshot = sub {
...
 
+    # remove snapshots from descendant subsystems that have the property "enabled" to "off", if the
+    # "recursive" flag is set to "on"
+    if ($backupSet->{recursive} eq 'on') {
+
+        $self->zLog->info("checking ZFS dependent datasets from '$backupSet->{src}' explicitely excluded");
+
+        # restrict the list to the datasets that are descendant from the current
+        my @dataSetList = grep /^$backupSet->{src}($|\/)/, @{$self->zZfs->listDataSets()};
+        if ( scalar @dataSetList > 0 ) {
```
oetiker: `if (@dataSetList)` is fine

lib/ZnapZend.pm
```diff
@@ -427,6 +427,42 @@ my $createSnapshot = sub {
...
+                # if the property does not exist, the command will just return. In this case,
+                # the value is implicit "on"
+                $prop = <$prop> || "on";
+                if ( substr($prop, 0, length('off')) eq 'off' ) {
+                    push(@dataSetsExplicitelyDisabled, $dataSet . '@' . $snapshotSuffix);
+                }
+            }
+
+            # remove the snapshots previously marked
+           if ( scalar @dataSetsExplicitelyDisabled > 0 ){
```
oetiker: `if (@dataSetsExplicielyDisabled)`

lib/ZnapZend.pm
flixman: I have added a chomp also, because the command returns with a '\n'.

flixman:
Seems the project has a build error with perl 5.26, but is an amazon s3 error :-S @oetiker ; I think I can do nothing to get this fixed?

oetiker:
Sorry I am on the road, but I get the impression that the error is from a missing module: Test::Exception ... adding this to PERL_MODULES could fix the problem

flixman:
But, AFAIK, I can not modify travis configuration, right? (Just to confirm whether I can or I cannot fix it)

oetiker:
you can modify everything ... just edit .travis.yml ... this is what gets used to test your PR

oetiker:
thanks

oetiker:
refers to #154

PR #380, opened 2018-09-24
Further fix for service integrations (fix evaluated vars other than PREFIX)

jimklimov:
Should fix inadequacies reported in #348 

Also updates `Makefile.in` for users who do not `./bootstrap.sh` by themselves.


oetiker:
thanks

PR #383, opened 2018-10-06
Notify when a property is not available, and skip to the next dataset.

flixman:
When a property on a dataset is not available, znapzend dies with an error message. This PR changes that by a warning and the skipping of that dataset. By doing this, we can set org.znapzend properties on descendant datasets (specifically thinking on the org.znapzend:enable, to skip those datasets that have it set to "off" when coming from a recursive parent).


oetiker:
sorry missed the context ... getting a warning every time around does not seem all that good a solution either ...

flixman:
I am running it right now, and I am only getting a warning when the service is reloaded or started (which is the only moment when the backup plan is refreshed).

Issue #385, opened 2018-10-09
Can speed up a lot with recursive "zfs destroy"


jimklimov:
The behavior I see currently in practice, and in code, is that after the snapshots are made and sent, they are destroyed - as one huge list of arguments by default or one by one in `oracleMode`. Our setup uses an extensive tree of datasets with a recursive znapzend policy starting from a low-hanging branch of the pool, so there are typically thousands of snaps to delete and processing each one takes the host several seconds for all the synchronous contexts... so the loop just does not keep up doing the job, and we run out of space due to obsolete snapshots eating it up with referenced unneeded bits.

Note 1: if I use backgrounded `zfs destroy` commands from a loop and/or a recursive destroy, it takes as long to destroy hundreds of snapshot datasets as it takes to do one, so there ought to be some common big lock there effectively caching and coalescing these requests.

Note 2: in fact the processing of blocks getting freed up is background and asynchronous on even later Solaris 10 releases and illumos nowadays, and can take minutes after the zfs CLI commands have completed and returned. But getting those blocks onto the hit-list takes a considerable while.

I see that creation of snapshots takes the ZFS recursion support into account at https://github.com/oetiker/znapzend/blob/c604a86857430258c2b8479c356437c0f61a4dc6/lib/ZnapZend/ZFS.pm#L221 but removal does not seem to: https://github.com/oetiker/znapzend/blob/c604a86857430258c2b8479c356437c0f61a4dc6/lib/ZnapZend/ZFS.pm#L235

My suggestion is to start the cleanup phase with a quick *recursive* deletion of the specified snapshot name from the root branch with an individual local znapzend zetup, especially if we know that it was making recursive snapshots in the first place. Then we can follow up with the existing logic to find possibly missed snapshots in child datasets that should also be removed. Hopefully this latter part would usually have nothing to do.

I'll try to give this idea a run in our deployments before PRing, but comments are welcome in general :)

jimklimov:
Seems the best place for the change is to first process `$backupSet->{src}` (rather than generalized `$srcSubDataSets` right away), using a `destroySnapshots()` extended with recursive flag support like used in `createSnapshots()`, around https://github.com/oetiker/znapzend/blob/4ccbe714186ac1fbc72a81e0548e6178279b8c76/lib/ZnapZend.pm#L354

[...]

PR #386, opened 2018-10-11
Recursive destroy

jimklimov:
This should take care of #385, but so far has not been tested on real systems ;)


jimklimov:
Tested on local system, wrinkled out some issues and optimized a bit, seems to work for me.

oetiker:
what would be really cool would be if you enhanced the test suite accordingly

jimklimov:
I am not actually sure how to go about auto-testing it at this time :\

I did `git grep destroySnapshot` and found one more use-case I missed before :) but otherwise there were no apparent hits under tests. A `gmake check` fails for me regardless - in the master branch as well...

jimklimov:
Bumped to align with current master.

Also read up today on ability to have resursive mode but certain datasets not enabled.

jimklimov:
Updated to align with current master. Every PR wants some love one year or another ;)

Also read today about an ability to have recursive backup plan, but have it not enabled on some child datasets. I wonder if this recursive mass-removal would contradict anything (e.g. cleaning up datasets that we did not intend to actively back up? good... To touch? bad...)

oetiker:
any thoughts on how to add testing ?

jimklimov:
Probably a few lines in random files under `t/` directory should do it. Just gotta carve them out :)

jimklimov:
So I somewhat cheated at first and covered the previously existing but not-covered code ;)

And went from there to cover some of the new paths as well.

jimklimov:
I anticipate that after merging #439, the coverage would also include large blocks about (cleaning) child datasets that inherit a backup plan but are not directly configured. I did not port it here because it largely depends on changes made for that PR, in both mock `zfs` and maybe the production code.

jimklimov:
Idea for after merging these PRs (and so taking advantage of the logical structure in that one) - also add `tank/source` child datasets that would have `org.znapzend:enabled==off` into `t/zfs` output, and so test -cover one more family of codepaths.

oetiker:
let me know when you deem the whole thing complete :) I will then review!

jimklimov:
I hope the two PRs are now complete (and beyond, with the testing cleanups) and should hopefully not conflict when merging :)

PR #395, opened 2018-11-04
Don't use deprecated Mojo::Util::slurp

mas90:
As of Mojo 7.15, Mojo::Util::slurp is deprecated in favour of Mojo::File::slurp. As of 7.31, it is removed entirely. We only use it to read the PID file.


oetiker:
good point

Issue #229, opened 2016-08-08
Direct mbuffer transfer working with Linux ?


mnbokaem:
znapzend works fine for me when using mbuffer=off or mbuffer=/usr/bin/mbuffer, essentially when running through ssh.
However when using mbuffer with the port option for direct communication, the underlying send/recv processes never get launched. The OS is Linux with ZoL.

Config with issues:

```
root@znas:/opt/znapzend-0.15.7# znapzendzetup list
*** backup plan: zfast/local/bamboo ***
dst_0           = root@qnas:qarchive/znas/local/bamboo
dst_0_plan      = 10days=>6hours,30days=>1day,90days=>1week
enabled         = on
mbuffer         = /usr/bin/mbuffer:2345
mbuffer_size    = 200M
post_znap_cmd   = off
pre_znap_cmd    = off
recursive       = off
src             = zfast/local/bamboo
src_plan        = 4days=>6hours
tsformat        = %Y-%m-%d-%H%M%S
```

Debug output with that config:

```
root@znas:/opt/znapzend-0.15.7# ./bin/znapzend --debug --runonce=zfast/local/bamboo
[Mon Aug  8 20:53:08 2016] [info] znapzend (PID=12904) starting up ...
[Mon Aug  8 20:53:08 2016] [info] refreshing backup plans...
[Mon Aug  8 20:53:09 2016] [info] found a valid backup plan for zfast/local/bamboo...
[Mon Aug  8 20:53:09 2016] [info] znapzend (PID=12904) initialized -- resuming normal operations.
[Mon Aug  8 20:53:09 2016] [debug] snapshot worker for zfast/local/bamboo spawned (12909)
[Mon Aug  8 20:53:09 2016] [info] creating snapshot on zfast/local/bamboo
# zfs snapshot zfast/local/bamboo@2016-08-08-205309
[Mon Aug  8 20:53:09 2016] [debug] snapshot worker for zfast/local/bamboo done (12909)
[Mon Aug  8 20:53:09 2016] [debug] send/receive worker for zfast/local/bamboo spawned (12911)
[Mon Aug  8 20:53:09 2016] [info] starting work on backupSet zfast/local/bamboo
[Mon Aug  8 20:53:09 2016] [debug] sending snapshots from zfast/local/bamboo to root@qnas:qarchive/znas/local/bamboo
[...]

PR #399, opened 2018-11-29
Fix For ProxMox 5.x

pbucher:
I'm not sure if this is the current way to fix this or not. I'm not a Perl programmer at all and this is mostly Greek to me. 

That said I've been using ZnapZend for ages without issue under OmniOS and when I moved to proxmox 5.x(aka Debian Stretch with a custom kernel) ZnapZend stopped working. It may have worked on 4.x but I made the switch just has 5.x was released so I never ran production on 4.x. 

Anyways, I traced the issue down to section of code under '#spawn event' never coming to life. The remote recv process fires up without issue but the local send command never happens and any debug statements inside $fc->on( spawn => sub { never get called.

My fix is to add the the one_tick() call. I've been running this for a year now with great results. I was using the .19 release branch and recently moved to head, which out issue. I tested head first without my fix and got the same results as before.

Thank you for making this great little script.

### DIFF
--- a/lib/ZnapZend/ZFS.pm
+++ b/lib/ZnapZend/ZFS.pm
@@ -11,7 +11,6 @@ has nodestroy       => sub { 1 };
 has oracleMode      => sub { 0 };
 has recvu           => sub { 0 };
 has compressed      => sub { 0 };
-has netcat          => sub { 1 };
 has rootExec        => sub { q{} };
 has sendDelay       => sub { 3 };
 has connectTimeout  => sub { 30 };
--- a/lib/ZnapZend/ZFS.pm
+++ b/lib/ZnapZend/ZFS.pm
@@ -11,6 +11,7 @@ has nodestroy       => sub { 1 };
 has oracleMode      => sub { 0 };
 has recvu           => sub { 0 };
 has compressed      => sub { 0 };
+has netcat          => sub { 1 };
 has rootExec        => sub { q{} };
 has sendDelay       => sub { 3 };
 has connectTimeout  => sub { 30 };
@@ -410,6 +411,8 @@ sub sendRecvSnapshots {
         );
         #start forkcall event loop
         $fc->ioloop->start if !$fc->ioloop->is_running;
+        #wake up event loop
+        $fc->ioloop->one_tick();
     }
     else {
         my @mbCmd = $mbuffer ne 'off' ? ([$mbuffer, @{$self->mbufferParam}, $mbufferSize]) : () ;


pbucher:
See Issue #229 for additional background

pbucher:
Are you using it with remote destinations? I've tried this on several different Proxmox 5 servers and have always had the issue that the local thread never wakes up and executes the local send command. The remote server gets the recv command and is running waiting for the send that never comes.

I build the package using [https://github.com/Gregy/znapzend-debian](url)

Also here is my znapzend config for reference:

> dst_a=root@zback:rpool/backupz/bbsc/ocean_rpool
dst_a_plan=1days=>15minutes,5days=>1days,1years=>1weeks
enabled=on
feature=sudo,compressed
mbuffer=/usr/bin/mbuffer:13871
mbuffer_size=2G
post_znap_cmd=off
pre_znap_cmd=off
recursive=on
src=rpool/data
src_plan=3days=>15minutes,30days=>1days
tsformat=%Y-%m-%d-%H%M%S

redmop:
I have 1 local to external USB HDD. It's been ok.

lib/ZnapZend/ZFS.pm
```diff
@@ -410,6 +410,8 @@ sub sendRecvSnapshots {
         );
         #start forkcall event loop
         $fc->ioloop->start if !$fc->ioloop->is_running;
+        #wake up event loop
+        $fc->ioloop->one_tick();
```
oetiker: hmmm this whole section looks a bit odd to me ... have you looked at the state of `$fc->ioloop->is_running` ?  Because if this is NOT true, then my guess is that the reason for the problem you are seeing is that $fc is going out of scope at the end of the block and thus the event handlers are gone too ... with the one_tick call you allow the spam to happen ... but then it would make sense to make the is_running call into a propper if/then condition.

Beneter:
`Mojo::IOLoop::ForkCall` was deprecated in favour of `Mojo::IOLoop::Subprocess`. If I had the time I would like to try if this issue still persists with the latter.

oetiker:
what we especially need is a way for the subprocess to report stderr output from ssh back to the master process so that logging works properly.

pbucher:
I'm happy to report the problem no longer exists on Proxmox 6.x. I'm not 100% sure what the root cause was with Promox 5.x.  If I ever get a few spare cycles maybe I'll give a shot at switching things over to Mojo::IOLoop::Subprocess.

PR #439, opened 2019-10-14
Znapzendzetup list recursive

jimklimov:
So I was debugging my laptop backup setup today, and was annoyed that I must either `znapzendzetup list` source datasets that I know to have been configured by name, one by one, or suffer the whole-system dataset listing (with USB HDD taking many minutes to give the list)...

And so I scratched a few itches:
* added ability to list several named dataset configs in one command call
* added ability to recurse into children of named dataset(s)
* added a fallback for systems whose memory vs ZFS tree sizing won't permit listing all the child dataset attributes :-)
* extended `ZFS.pm` routines `getDataSetProperties()` and `listDataSets()` with ability to inspect not only a single string `$dataSet` value, but also arrays of strings, to reduce the number of (slow) calls into `zfs` command
* added optional debugging and features (for lowmemRecurse) support in `znapzendzetup`

Tested with various combinations locally, seems to work for stuff I typed :)

But CI selftests not yet updated, probably `zfs` mockup needs some love :(


lib/ZnapZend/ZFS.pm
```diff
@@ -431,28 +456,104 @@ sub sendRecvSnapshots {
...
+    # as rarely as we can and so complete faster. However, on systems
+    # with too many datasets this can exhaust the memory available to
+    # the process, and/or time out. To defend against this, we optionally
+    # can fall back to a big list of individual dataset names found by
+    # recursive listDataSets() invocations instead.
+    if ($dataSet) {
+        if (ref($dataSet) eq 'ARRAY') {
+            print STDERR "=== getDataSetProperties(): Is array...\n" if $self->debug;
+            if ($self->lowmemRecurse && $recurse) {
+                my $listds = $self->listDataSets(undef, $dataSet, $recurse);
```
jimklimov: Also a question for reviewers : in the fallback logic here and a bit below, I pass `undef` for the `$remote` meaning to look at the local system. First it seemed to match the earlier call when used without any arguments, or verbatim use of the `$dataSet` string when with arguments.

However here we can pass several args (as an array), some of which potentially could have the `user@remote:data/set` format... should I bother about this possibility now? :-) I do not think this change breaks any use-case fundamentally, but maybe others have more visibility and memory :)

jimklimov:
Oh wow, seeing green CI builds now :)

jimklimov:
Can't make sense of some logs... says tests are green but by coverage seems something failed mid-way.

And also this:

````
# zfs destroy remote/destination@2019-10-14-184000,2019-10-14-184200,2019-10-14-184300,2019-10-14-184400,2019-10-14-184500,2019-10-14-184600,2019-10-14-184700,2019-10-14-184800,2019-10-14-184900,2019-10-14-185100,2019-10-14-185200,2019-10-14-185300,2019-10-14-185400,2019-10-14-185500,2019-10-14-185600,2019-10-14-185700,2019-10-14-185800,2019-10-14-185900,2019-10-14-190200,2019-10-14-190300,2019-10-14-190400,2019-10-14-190600,2019-10-14-190700,2019-10-14-190800,2019-10-14-190900,2019-10-14-191100,2019-10-14-191200,2019-10-14-191300,2019-10-14-191400,2019-10-14-191600,2019-10-14-191700,2019-10-14-191800,2019-10-14-191900

# zfs list -H -o name -t filesystem,volume backup/destfail

[Mon Oct 14 19:40:13 2019] [warn] destination 'backup/destfail' does not exist or is offline. ignoring it for this round...
[Mon Oct 14 19:40:13 2019] [warn] ERROR: suspending cleanup source dataset because at least one send task failed
[Mon Oct 14 19:40:13 2019] [info] done with backupset tank/source in 3 seconds
[Mon Oct 14 19:40:13 2019] [info] znapzend (PID=6792) initialized -- resuming normal operations.
not ok 4 - znapzend dies with no backup sets defined or enabled at startup
#   Failed test 'znapzend dies with no backup sets defined or enabled at startup'
#   at ./t/znapzend.t line 69.
# expecting: Regexp ((?^:No backup set defined or enabled))
# found: normal exit
````

So was it supposed to fail but didn't (and have I broken something?) or not?

**Comment by jimklimov** [comment]
Seems I did... 

````
=== getDataSetProperties(): Is array...
# zfs list -H -o name -t filesystem,volume -r tank/source
cannot open 'tank/source': dataset does not exist
=== getDataSetProperties(): List all...
# zfs list -H -o name -t filesystem,volume
````

at least somtehing

**Comment by jimklimov** [comment]
Seems the CI gods are satisfied now... though in the build log I see
````
ok 15 - znapzendzetup import --write
# Tests were run but no plan was declared and done_testing() was not seen.
# Looks like your test exited with 1 just after 15.
````

and locally the test 16 passes:

````
ok 15 - znapzendzetup import --write
=== getDataSetProperties():
...
ok 16 - znapzendzetup list --features=lowmemRecurse --debug --recursive tank/source
1..16
````

File: lib/ZnapZend/ZFS.pm [hunk 2]
```diff
@@ -431,28 +456,129 @@ sub sendRecvSnapshots {
...
+        }
+        push (@list, @{$listds});
+    }
+
+    for my $listElem (@list){
+        print STDERR "=== getDataSetProperties(): Looking under '$listElem' with '$recurse' recursion mode\n" if $self->debug;
         my %properties;
-        my @cmd = (@{$self->priv}, qw(zfs get -H -s local -o), 'property,value', 'all', $listElem);
+        my @cmd = (@{$self->priv}, qw(zfs get -H -s local));
+        push (@cmd, qw(-t), 'filesystem,volume');
```
jimklimov: Also for reviewers: did any users of the older implementation of `getDataSetProperties()` rely on its ability to query non-fs/vol datasets (e.g. ask for properties of snapshots, or of whatever future dataset types may be defined)?

It can be added back for non-recursive case of explicitly listed dataset(s) to query, but it explodes size- and time-wise if we recurse a pool with hundreds of thousands of snapshots and get attributes of those too.

oetiker: I am not aware of such a use ...

lib/ZnapZend/ZFS.pm
oetiker: I would suggest to have as little ssh magic on the commandline and rather encurrage people to create appropriate .ssh/config files and then work with `remote` names matching these definitions.

oetiker:
make sure to update the test suit for your new features.

jimklimov:
Yes, on it. Essentially, there is one "real" codepath that is not called and I'm fixing (the recursion over several datasets listed by `zfs` in one shot), and the rest ov coverage misses are mostly error-cases that we do not trigger. I'll try to add some faults :)

lib/ZnapZend/Config.pm
```diff
@@ -25,7 +26,7 @@ has mandProperties => sub {
     }
 };
 
-has zfs  => sub { my $self = shift; ZnapZend::ZFS->new(rootExec => $self->rootExec); };
+has zfs  => sub { my $self = shift; ZnapZend::ZFS->new(rootExec => $self->rootExec, debug => $self->debug, lowmemRecurse => $self->lowmemRecurse); };
```
oetiker:
```suggestion
has zfs  => sub { 
    my $self = shift;
    ZnapZend::ZFS->new(
        rootExec => $self->rootExec,
        debug => $self->debug, 
        lowmemRecurse => $self->lowmemRecurse
    ); 
};
```

lib/ZnapZend/Config.pm
```diff
@@ -156,18 +157,38 @@ my $checkBackupSets = sub {
...
+    #get all backup sets and check if valid, from remainder of ARGV
+    if (scalar(@_) > 0) {
+        $self->backupSets($self->zfs->getDataSetProperties(\@_, $recurse));
+    } else {
+        # Not that recursion makes much sense for "undef" (=> list everything)
+        $self->backupSets($self->zfs->getDataSetProperties(undef, $recurse));
+    }
     $self->$checkBackupSets();
 
+    printf STDERR "=== getBackupSet() : got " . scalar(@{$self->backupSets}) . " dataset(s) with a local backup plan\n" if $self->debug;
```
oetiker:
```suggestion
    printf STDERR "=== getBackupSet() : got " 
        . scalar(@{$self->backupSets}) . " dataset(s) with a local backup plan\n" 
            if $self->debug;
```

lib/ZnapZend/Config.pm
```diff
@@ -156,18 +157,38 @@ my $checkBackupSets = sub {
...
+        return 0; # false
+    }
+
     if ($enabledOnly){
         my @backupSets;
 
         for my $backupSet (@{$self->backupSets}){
             push @backupSets, $backupSet if $backupSet->{enabled} eq 'on';
         }
+        printf STDERR "=== getBackupSet() : got " . scalar(\@backupSets) . " enabled-only dataset(s) with a local backup plan\n" if $self->debug;
```
oetiker:
```suggestion
        printf STDERR "=== getBackupSet() : got " 
            . scalar(\@backupSets) . " enabled-only dataset(s) with a local backup plan\n" 
                if $self->debug;
```

lib/ZnapZend/Config.pm
```diff
@@ -156,18 +157,38 @@ my $checkBackupSets = sub {
...
+        if (scalar(\@backupSets) == 0) {
```
oetiker:
```suggestion
        if (not @backupSets == 0) {
```

lib/ZnapZend/ZFS.pm
```diff
@@ -431,28 +456,129 @@ sub sendRecvSnapshots {
...
     my $self = shift;
     my $dataSet = shift;
+    my $recurse = shift; # May be not passed => undef
+
     my @propertyList;
     my $propertyPrefix = $self->propertyPrefix;
 
-    my $list = $dataSet ? [ ($dataSet) ] : $self->listDataSets();
+    my @list;
+    print STDERR "=== getDataSetProperties():\n\trecurse=" . Dumper($recurse) . "\n\tDS=" . Dumper($dataSet) . "\n\tlowmemRecurse=" . $self->lowmemRecurse . "\n" if $self->debug;
```
oetiker:
```suggestion
    print STDERR "=== getDataSetProperties():\n\trecurse=" 
        . Dumper($recurse) . "\n\tDS=" . Dumper($dataSet) 
        . "\n\tlowmemRecurse=" . $self->lowmemRecurse . "\n"
             if $self->debug;
```

lib/ZnapZend/Config.pm
oetiker: \@backupSets is a pointer to an array ... scalar of a pointer to an array would always be 1

jimklimov: Curiously, the code did work, bot for the `print` above (returning 0 or 1 or 5 entries as expected) and for the conditional here.

jimklimov:
Thanks for that other merge, glued it to this one now, waiting for re-testing :)

lib/ZnapZend/Config.pm
jimklimov: Oh, so maybe I did get the context wrong - this is the lower check, for enabledOnly stuff.

jimklimov: Posted https://github.com/oetiker/znapzend/pull/439/commits/8b20412a36edffd154feb9306c00a7ed197af8fa

PR #442, opened 2019-10-20
Problem: no easy way to runonce a pool or dataset tree

jimklimov:
Solution: add recurse ability to `znapzend --runonce=x -r` with same
logic as in `znapzendzetup list -r x`

PS: While CI did not complain, seems this also fixes an omission from my recent PR #439 about `getBackupSetEnabled()` usage without a recursion arg that it still should send into `getBackupSet()`.


jimklimov:
Oh, typos.

jimklimov:
Missed it by a tiny mark =)

`Coverage decreased (-0.01%) to 88.402%`

jimklimov:
Take this, Coveralls! :) `Coverage increased (+0.1%) to 88.526% `

PR #443, opened 2019-10-21
Docs and comments cleanup

jimklimov:
Make the playground tidier, so other PRs can concentrate just on their improvements :)

### DIFF
--- a/doc/znapzend.pod
+++ b/doc/znapzend.pod
@@ -1,4 +1,4 @@
-=head1 NAME
+[01;31m[K=head1 NAME[m[K
 
 znapzend - znapzend daemon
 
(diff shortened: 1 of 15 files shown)


doc/znapzend.pod
```diff
@@ -1,4 +1,4 @@
-=head1 NAME
+[01;31m[K=head1 NAME[m[K
```
oetiker:
```suggestion
=head1 NAME
```

doc/znapzendzetup.pod
```diff
@@ -1,4 +1,4 @@
-=head1 NAME
+[01;31m[K=head1 NAME[m[K
```
oetiker:
```suggestion
=head1 NAME
```

doc/znapzendztatz.pod
```diff
@@ -1,4 +1,4 @@
-=head1 NAME
+[01;31m[K=head1 NAME[m[K
```
oetiker: =head1 NAME

doc/znapzend.pod
jimklimov: Hmm, different tooling is differently eager it seems :)

PR #444, opened 2019-10-22
Znapzend inherited backup plans

jimklimov:
One big annoyance is when some small ZFS subtree has not replicated, but to retry getting it there I must `runonce` and wait for the whole backup config to be processed.

This PR adds `--inherit` flag to `znapzend` and `znapzendzetup list` to permit (non-default) use of datasets that have the `org.znapzend:*` attributes inherited, but only if it comes from a dataset whose definition source for them is "local". This way the first encountered (directly specified or recursed) dataset with a backup plan (local or such inherited) in a subtree would be the replication source (datasets under it that only have the znapzend attrs inherited are not added to explicit lists for processing of them with children).

The PR is currently posted as Work-in-progress to solicit early comments about overall sanity ;) and because it is late night now, and I likely missed something in the code and/or did not implement some nuances as "promised" by the documented plan in comments and help/man text.


bin/znapzend
```diff
@@ -164,6 +165,23 @@ just look at this dataset's backup plan, but iterate into its children that
...
+
+when backing up a particular dataset with B<--runonce>=[I<dataset>], do not
+require that the provided dataset directly has a locally defined backup plan
+configuration (sanity check that it is not "received" via ZFS replication).
+With this option in place, allow also to quickly snapshot and/or replicate
+datasets of subtrees which inherit a backup plan from a dataset which has it
+in its attributes whose source is "local".
+
+At this time, the B<--inherited> is effectively exclusive vs. B<--recursive>
+mode (and preferred if both are specified) if the requested dataset has some
```
oetiker: if there is exclusivity I would raise an error if both are specified.

bin/znapzend
```diff
@@ -164,6 +165,23 @@ just look at this dataset's backup plan, but iterate into its children that
...
+(local or inherited) backup plan. If a dataset without a backup plan is passed,
+the script should ignore the B<--inherited> option and just recurse until it
+finds some dataset(s) with a proper locally defined backup plan.
+
```
oetiker: adding an example use would be great to make the intended usage clear

bin/znapzend
jimklimov: Good point, I'll think how to re-word it. The idea was that de-facto one takes place, if both are set it depends on the explicitly requested dataset if it has a config, or in recursive mode the nearest dataset(s) under it having a config (likely local in this case... not sure if inherited is then even possible). So it would not recurse further, at least not into inherited-only configs,

One point to review is whether znapzend allows having another "local" backup plan config in a descendant dataset under one that has a config already.

bin/znapzend
jimklimov: By the way, is there a way in this test suite to check what the calls actually did, not qust that they succeeded or failed with specific exit code? (Maybe grep stdout+stderr)

Here I can set up a few scenarios where I expect that e.g. these 1-2-3 configs are found (amount and names), but do not really know how to check for it programmatically.

bin/znapzend
jimklimov: The documentation comments are hopefully addressed by "Revise doc...." commit now, thanks.

bin/znapzend
jimklimov: The documentation comments (for examples) are hopefully addressed by "Revise doc...." commit now, thanks.

bin/znapzend
oetiker: having a separate backup plan underneath one that is marked as recursive is something @moetiker just talked to me yesterday ... at the moment this is not possible, but he would sure like it :)

oetiker: one of the reasons why we wanted recursive backups was to have consistant snapshots over a tree, so if there was a alternate plan somewhere down the tree that would make the snapshotting and espeacially the cleanup a bit more involved. although you might just ignore the 'local snapshot config' for the this case ... complaining about it if someone configured it.

bin/znapzend
oetiker: re testing ... the test suit is in perl and in shell, so yes anythihng is possible :) if you have a concrete example I may be able to advise

jimklimov:
I think by now it does what I intended so want to drop the "WIP" tag and let you merge it :-) Though maybe more people giving it a run first and finding any unfortunate regressions would also help.

As commented in the code, there are some nuances that could be handled better (e.g. more optimal processing/tracking of sub-inherited datasets we would ignore) but these refinements might be in some other PR.

PR #445, opened 2019-10-22
Problem: selftests spew "Variable will not stay shared"

jimklimov:
Solution: change file-scoped private functions and methods from "my" to "our"

Note for reviewers: I am really not sure if this scoping keywork change would have any adverse effects. A quick local run of self-tests does not seem to expose any issues, let's see what Travis would say.

### DIFF
--- a/lib/ZnapZend.pm
+++ b/lib/ZnapZend.pm
@@ -99,7 +99,8 @@ has zLog => sub {
     return $log;
 };
 
-my $killThemAll  = sub {
+### private methods ###
+our $killThemAll  = sub {
     my $self = shift;
 
     $self->zLog->info("terminating znapzend (PID=$$) ...");
@@ -124,7 +125,7 @@ my $killThemAll  = sub {
     exit 0;
 };
 
-my $refreshBackupPlans = sub {
+our $refreshBackupPlans = sub {
     my $self = shift;
     my $recurse = shift;
     my $dataSet = shift;
@@ -213,7 +214,7 @@ my $refreshBackupPlans = sub {
     }
 };
 
-my $sendRecvCleanup = sub {
+our $sendRecvCleanup = sub {
     my $self = shift;
     my $backupSet = shift;
     my $timeStamp = shift;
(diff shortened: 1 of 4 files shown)


jimklimov:
Note: Coverage changed (grew) because some single-line monsters were reformatted to multiple lines :)

lib/ZnapZend.pm
```diff
@@ -99,7 +99,8 @@ has zLog => sub {
     return $log;
 };
 
-my $killThemAll  = sub {
+### private methods ###
+our $killThemAll  = sub {
```
oetiker: The point of using `my` is to 'hide' the method from external access. By using `our` you make the methods publicly accessible ...  so you may want to do these changes just in the code that modifies the source at runtime before executing the tests.

jimklimov: Huh, is there some such feature? :-)

Could you please close this PR and do it the right way then, I don't know how to approach it this way properly and am getting bogged on other PRs and work that I do have ideas about :)

oetiker: in the test code there is a function which modifies the perl code prior to runing the tests ... these modifications are the root of the warnings ... that is why they only occur while runnig the tests ... so the solution you found would be fine but it has to be at runtime in this code ... https://github.com/jimklimov/znapzend/blob/unshared-vars/t/znapzend.t#L20

jimklimov: Oh wow :) I thought it was due to the `use strict; use warnings;` which are not used in production code under `bin/` but are enforced by import into test scripts.

jimklimov: Hopefully got it done :)

jimklimov:
FWIW, the reduced coverage is due to added optional writing of rewritten perl code to help trace the process. Tested locally that

````
DEBUG_ZNAPZEND_SELFTEST_REWRITE=yes ./t/znapzend.t 
````

works to create these temporary outputs (properly git-ignored), and that without the envvar these code dumps do not appear in the filesystem.

Thanks for the hints and a new trick learned ;)

PR #450, opened 2019-11-06
fix znapzend dataset discovery without arguments

lotheac:
In master, `znapzend` without arguments fails as follows:
```
% znapzend -d
=== getDataSetProperties():
        recurse=$VAR1 = 1;

        inherit=$VAR1 = 0;

        DS=$VAR1 = [
          ''
        ];

        lowmemRecurse=0
=== getDataSetProperties(): Is array...
=== getDataSetProperties(): Looking under '' with '1' recursion mode and '0' inheritance mode
# zfs get -H -s local -t filesystem,volume -r -o name,property,value,source all
cannot open '': empty component in name
=== getDataSetProperties():
        Collected:
=== getBackupSet() : got 0 dataset(s) with a local backup plan
No backup set defined or enabled, yet. run 'znapzendzetup' to setup znapzend
```

this happens because the zfs command is passed an empty string as an extra argument (see DS= in the debug output). So make sure that doesn't happen by default.


jimklimov:
Curious, I'm sure I tested this at some point in development... maybe not at the latest point though. Sorry for the mess-up and thanks for taking care of it :-)

PR #453, opened 2019-11-14
Fix backup replication of inherited-config datasets

jimklimov:
Before this, they were trying to replicate into the original (parent) `dst_N` rather than into a properly shifted one.

### DIFF
--- a/lib/ZnapZend/ZFS.pm
+++ b/lib/ZnapZend/ZFS.pm
@@ -740,7 +740,23 @@ sub getDataSetProperties {
                     }
                     if (defined($cachedInheritance{$tail_inheritKey}) && 1 == $cachedInheritance{$tail_inheritKey}) {
                         # This attr comes from a local source
-                        $properties{$key} = $value;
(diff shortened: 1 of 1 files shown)


jimklimov:
Note: one failed line of coverage is for the case I am not sure we can hit in current reality, but handle for consistency just in case.

jimklimov:
@oetiker : I think this one is good to merge :)

oetiker:
if you request a review, then I know you think the PR is ready ... just click my name in the reviewer list.

jimklimov:
As not a project member, I don't have a link to click in the Reviewers column ;) So have to CC in comments...

jimklimov:
Seems good, no `not ok` and `done_` hits found in Travis build logs :)

jimklimov:
The coverage fault is the same case as above, for displaying an error in the situation we don't really expect to happen IRL.

jimklimov:
Green and covered (with the conditional-miss above), is there something more to fix here? :)

PR #458, opened 2019-11-22
Fix --runonce mode with no dataset arg

jimklimov:
Stumbled on this discrepancy of doc vs code vs tests and mocks while extending selftests for #457


jimklimov:
Seems good, no `not ok` and `done_` hits found in Travis build logs :)

PR #459, opened 2019-11-22
Introduce a skipIntermediate feature for zfs sending faster and smaller chunks

jimklimov:
Follows up from @lotheac 's #455 and https://github.com/lotheac/znapzend/pull/1


jimklimov:
@oetiker : This one seems good, no `not ok` and `done_` hits found in Travis build logs :)

PR #461, opened 2019-11-22
Small fixes for enable-dst/disable-dst

jimklimov:


jimklimov:
Green and covered, is there something more to fix here? :)

PR #468, opened 2020-01-25
Add sendRaw as a feature to enable sending encrypted datasets without decrypting

pspikings:
Added as a feature in the same way as compressed and recvu. I think it might be better to have this feature defined with destinations and stored in the dataset attributes (so when sending to two remotes, one could be sent encrypted and the other not) but that applies to compressed and recvu as well so a bigger piece of work!


oetiker:
with this option all destinations get the raw treatment ... what is the thinking behind this approach ?

PR #475, opened 2020-02-20
Do not back up from bogus snapshot datasets

jimklimov:
Hit an issue on a system that had some mid-way state of codebase (where I developed other znapzend patches) that it somehow listed a snapshot into the array of datasets to back up. This failed when requesting its properties (it is not a "filesystem,volume" type for zfs call) and aborted the backup.

The original call was with non-mainstream mix of "runonce recursive inherited" so maybe there is some bug there that I did not yet look at further. In fact maybe it was some bug in the older code version abandoned on that system that may be no longer present in committed sources - it is complicated to check better because that box is away from internet and does not have git tools.

Anyway, we can alert the users if their code ever gets into this state (so the bug can be confirmed if it happens in the wild), and skip the snapshots from the list they should not be in.

Note: First I considered adding the is/is-not a snapshot checks based on characters in `$dataSet` string into `datasetExists()` and many others near it, but chose against it. Let ZFS of the host's version decide generally what is or is not a requested dataset type, now that there are also bookmarks etc. in the mix.

### DIFF
--- a/lib/ZnapZend/Config.pm
+++ b/lib/ZnapZend/Config.pm
@@ -89,6 +89,17 @@ my $checkBackupSets = sub {
            next;
         }
 
+        if ( $dataSet =~ m/[\@]/ ) {
+            # If we are here, somebody fed us a snapshot in the list of
+            # datasets, which is likely a bug elsewhere in discovery.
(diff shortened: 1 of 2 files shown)


oetiker:
thanks

PR #478, opened 2020-02-27
Set up spell-checking action

jsoref:
In case you'd like to enforce spell checking on commits/PRs, you could do this.

(As configured, it relies on a second branch to retain the whitelist/exclusions, but you can include that metadata inside the main branch -- just have the exclusion list exclude the two files)


oetiker:
nice ... you may well include it in the same branche

jsoref:
I should add one caveat: checks don't work terribly well when the PR is from a foreign repository because they don't have enough permission to comment.

If the person who created the PR has configured actions to run in forks, then they'll have already received the notification in their fork when they pushed before they create the PR.

To address that, there's an extra action `schedule` which tries to loop through open PRs and perform the checking (it tries to do it only once per pr update per window).

What will happen if you don't have the schedule/pr catching things is you'll get annotations on master and notifications and then you can fix it there.

I'll add the schedule config bit, probably over the weekend.

PR #486, opened 2020-05-05
Fixes some undefined values

samoser:


lib/ZnapZend.pm
```diff
@@ -356,21 +356,20 @@ my $sendRecvCleanup = sub {
 
             $self->zLog->debug('sending snapshots from ' . $srcDataSet . ' to ' . $dstDataSet);
             {
-                local $@;
```
oetiker, line 359: do NOT drop this ... it is good practice to set it!

Issue #495, opened 2020-07-22
Snapshot send encrypted to a backup destination for the first time fails


hansoostendorp:
Current behaviour:
Run ZnapZend on an encrypted source dataset, using the sendRaw feature to keep the data encrypted, on the destination dataset for the first time.
Result when encrypted destination dataset doesn't exist yet: send error, a non encrypted empty destination dataset is created.
Result when empty encrypted destination dataset exist: send error, destination dataset remains empty.

Expected behaviour:
The snapshot that is created by ZnapZend on the encrypted source dataset, is send to the destination dataset without errors and is still encrypted.

Work around:
After manually sending the first snapshot encrypted to the destination dataset, following snapshots created by ZnapZend are send encrypted successfully.

OS: Ubuntu 20.04 LTS
ZFS: 0.8.3
ZnapZend: 0.20.0

How te reproduce:

Create encrypted datapool/test source dataset
`# sudo zfs create -o encryption=on -o keyformat=passphrase datapool/test`

Create a backup plan for datapool/test to backuppool/test
`# sudo znapzendzetup create SRC '1h=>1min' datapool/test DST:a '1h=>1min' backuppool/test `

Run ZnapZend, using the sendRaw feature to keep the backup encrypted, and let ZnapZend automatically create the backup destination
`# sudo znapzend --debug --runonce=datapool/test --features=sendRaw --autoCreation`

It will automatically create the destination dataset using this zfs command
`# zfs create -p backuppool/test`

It will then run this send command resulting in an error
```
# zfs send -w datapool/test@2020-07-22-121805|zfs recv -F backuppool/test
cannot receive new filesystem stream: zfs receive -F cannot be used to destroy an encrypted filesystem or overwrite an unencrypted one with an encrypted one
```

The destination dataset that is created by ZnapZend is not encrypted and empty. When I run ZnapZend without autoCreation, it complains the destination dataset doesn't exist. When I create an encrypted destination dataset by hand, and run ZnapZend without the autoCreation option, the same error occurs and the destination dataset is still empty.

I think this is related to this ZendZnap issue: #302 
Which is related to this ZFS issue: [zfs receive -F cannot be used to destroy an encrypted filesystem #6793](https://github.com/openzfs/zfs/issues/6793)

What I take from these posts, is that the -F flag is only required if you have existing data to overwrite. When running the backup for the first time, there is no existing data, so the -F flag should be omitted.

[...]

Issue #491, opened 2020-06-30
Raw (encrypted) send failure


zdykstra:
When attempting to use `znapzend` to send a snapshot on a pool that's fully encrypted to a remote pool that is unencrypted, I'm receiving an error, even though the snapshot(s) do arrive on the destination without issue. 

**Sender details:** Void Linux, ZFS 0.8.4
```
# zfs get encryptionroot,encryption,keyformat,keylocation,keystatus zroot
NAME   PROPERTY        VALUE                      SOURCE
zroot  encryptionroot  zroot                      -
zroot  encryption      aes-256-gcm                -
zroot  keyformat       passphrase                 -
zroot  keylocation     file:///etc/zfs/zroot.key  local
zroot  keystatus       available                  -

# zfs list
NAME                         USED  AVAIL     REFER  MOUNTPOINT
zroot                       28.4G   433G      192K  none
zroot/ROOT                  16.0G   433G      192K  none
zroot/ROOT/void.2020.06.22  8.24G   433G     8.24G  /
zroot/home                  12.4G   433G     12.4G  /home

# zfs list -t snapshot
no datasets available

# znapzendzetup list
*** backup plan: zroot/ROOT ***
           dst_0 = root@rethe:zroot/backups/auri
      dst_0_plan = 60days=>4hours
         enabled = on
         mbuffer = off
    mbuffer_size = 1G
   post_znap_cmd = off
    pre_znap_cmd = off
       recursive = on
             src = zroot/ROOT
        src_plan = 30days=>4hours
        tsformat = %Y-%m-%d-%H%M%S
      zend_delay = 0
```
**Receiver details:** Void Linux, ZFS 0.8.4. For the purposes of this test, `zroot/backups` has been destroyed, so we're starting from scratch. 

**Full log of a run**
```
# znapzend --loglevel=debug --features=sendRaw --autoCreation --runonce=zroot/ROOT
[2020-06-29 21:11:22.78305] [23015] [info] znapzend (PID=23015) starting up ...
[2020-06-29 21:11:22.78330] [23015] [info] refreshing backup plans for dataset "zroot/ROOT" ...
cannot open 'zroot/backups/auri': dataset does not exist
cannot open 'zroot/backups/auri': dataset does not exist
[2020-06-29 21:11:24.62637] [23015] [info] creating destination dataset 'root@rethe:zroot/backups/auri'...
[2020-06-29 21:11:24.62740] [23015] [info] found a valid backup plan for zroot/ROOT...
[...]

PR #496, opened 2020-07-25
Support --autoCreation for sendRaw feature

hansoostendorp:
This change addresses these issues:
- Snapshot send encrypted to a backup destination for the first time fails #495
- Raw (encrypted) send failure #491

The issue is the way how znapzend handles destinations that do not exist. This isn't compatible with how ZFS handles a raw encrypted stream. This behaviour is introduced in PL #468 when the sendRaw feature is introduced. ZnapZend always uses the -F receive option that cannot be used to destroy an encrypted filesystem or overwrite an unencrypted one with an encrypted one. And ZnapZend creates an unencrypted destination dataset when --autoCreation is used. This triggers a ZFS error when sending a snapshot using a raw encrypted stream when the destination doesn't exist.

Changes in this PL:
- Do not use the -F receive option when sendRaw feature is used
- Do not create an unencrypted destination dataset when the sendRaw feature and--autoCreation is used, but let ZFS automatically take care of this.
- Do not send snapshots to the destination when the sendRaw feature is used and --autoCreation is omitted and the destination dataset doesn't exist.

This way the behaviour of --autoCreation is the same for normal streams and raw encrypted streams, and all information displayed to the user about destinations that do not exist and how to address this are still valid.

### DIFF
--- a/lib/ZnapZend.pm
+++ b/lib/ZnapZend.pm
@@ -317,12 +317,12 @@ my $sendRecvCleanup = sub {
         }
 
         #recheck non valid dst as it might be online, now
-        if (!$backupSet->{"dst_$key" . '_valid'} && !$self->sendRaw) {
+        if (!$backupSet->{"dst_$key" . '_valid'}) {
 
             $backupSet->{"dst_$key" . '_valid'} =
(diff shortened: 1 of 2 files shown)


hansoostendorp:
Thanks!

PR #497, opened 2020-07-28
Runonce --since=X made safer (only do dangerous stuff if --sinceForced=X is asked)

jimklimov:
This PR extends the proposal from #492 with ideas for failsafes and a more deterministic behavior that can be ordered by the user e.g. to enforce or not removal of destination snapshots newer than the one specified in `--since=X` argument to make it appear in the destination (for that effect, adds `--sinceForced=X` CLI option), and generally for other maintenance usages a `--forbidDestRollback` CLI option.

NOTE: at the time of this posting this is completely not tested in practice, a "theoretical PoC" code for discussion, so separate from the original PR adding `--since=X`.


jimklimov:
Rebased over recent master...

jimklimov:
Experimental data point:
* In relation to #503 I checked that sending with either `--since` or `--sinceForced` of a dataset that has *completely* no snapshots in common between source and destination (neither znapzend, nor others) refused to replicate, rollback, clean up etc. and so potentially cause loss of data. Reasonably, that is for me to do manually (to resync my rootfs clones).

jimklimov:
Rebased over recent master again, added small bugfixes and generated manpage updates...

jimklimov:
After the fix (notably, to look at snapshots not other datasets)...

* Sending with a now softer `--since=2020-04-14-09:55:32` which does not exist on source... but there are other datasets shared in history: user is non-fatally warned, and replication happens:
````
/bin/znapzend --features=zfsGetType,recvu,compressed,oracleMode --autoCreation --nodestroy --runonce=nvpool/ROOT/hipster_2019.10-20200127T101549Z --inherited --debug --since=@2020-04-14-09:55:32
The --since option is used and will try to ensure that the snapshot exists in history of destinations only if an older snapshot is the latest common (would not delete and rewrite subsequent snapshots to make way) at /usr/src/znapzend/bin/znapzend line 96.

...
[2020-07-31 13:55:51.97800] [8480] [debug] sending snapshots from nvpool/ROOT/hipster_2019.10-20200127T101549Z to backup-adata/snapshots/nvpool/ROOT/hipster_2019.10-20200127T101549Z
[2020-07-31 13:55:51.97841] [8480] [debug] Are we sending "--since"? since=="2020-04-14-09:55:32", skipIntermediates=="0", forbidDestRollback=="1"
# zfs list -H -o name -t snapshot nvpool/ROOT/hipster_2019.10-20200127T101549Z@2020-04-14-09:55:32
cannot open 'nvpool/ROOT/hipster_2019.10-20200127T101549Z@2020-04-14-09:55:32': dataset does not exist
### [--since mode]: Source dataset nvpool/ROOT/hipster_2019.10-20200127T101549Z does not have a snapshot named by --since='2020-04-14-09:55:32' at /usr/share/src/znapzend/bin/../lib/ZnapZend.pm line 485.
# zfs list -H -o name -t snapshot -s creation -d 1 nvpool/ROOT/hipster_2019.10-20200127T101549Z
# zfs list -H -o name -t snapshot -s creation -d 1 backup-adata/snapshots/nvpool/ROOT/hipster_2019.10-20200127T101549Z
# zfs send -Lce -I 'nvpool/ROOT/hipster_2019.10-20200127T101549Z@znapzend-auto-2020-07-31T11:49:09Z' 'nvpool/ROOT/hipster_2019.10-20200127T101549Z@znapzend-auto-2020-07-31T11:55:37Z'|/usr/bin/amd64/mbuffer -q -s 256k -W 600 -m 1G|zfs recv -u 'backup-adata/snapshots/nvpool/ROOT/hipster_2019.10-20200127T101549Z'
````
* same done with `--sinceForced` now fails to fulfill the user's harder requirement:
````
:; ./bin/znapzend --features=zfsGetType,recvu,compressed,oracleMode --autoCreation --nodestroy --runonce=nvpool/ROOT/hipster_2019.10-20200127T101549Z --inherited --debug --sinceForced=@2020-04-14-09:55:32
The --sinceForced option is used and will try to ensure that the snapshot exists in history of destinations (can delete and rewrite subsequent snapshots) at /usr/src/znapzend/bin/znapzend line 92.
...
[2020-07-31 14:05:39.35143] [8715] [debug] sending snapshots from nvpool/ROOT/hipster_2019.10-20200127T101549Z to backup-adata/snapshots/nvpool/ROOT/hipster_2019.10-20200127T101549Z
[2020-07-31 14:05:39.35192] [8715] [debug] Are we sending "--since"? since=="2020-04-14-09:55:32", skipIntermediates=="0", forbidDestRollback=="0"
# zfs list -H -o name -t snapshot nvpool/ROOT/hipster_2019.10-20200127T101549Z@2020-04-14-09:55:32
cannot open 'nvpool/ROOT/hipster_2019.10-20200127T101549Z@2020-04-14-09:55:32': dataset does not exist
### [--since mode]: Source dataset nvpool/ROOT/hipster_2019.10-20200127T101549Z does not have a snapshot named by --since='2020-04-14-09:55:32' at /usr/share/src/znapzend/bin/../lib/ZnapZend.pm line 485.
[2020-07-31 14:05:39.36427] [8715] [warn] User required --sinceForced='2020-04-14-09:55:32' but there is no match in source dataset nvpool/ROOT/hipster_2019.10-20200127T101549Z at /usr/share/src/znapzend/bin/../lib/ZnapZend.pm line 487.
...
[2020-07-31 14:05:39.43284] [8715] [warn] ERROR: suspending cleanup source dataset because 7 send task(s) failed:
[2020-07-31 14:05:39.43296] [8715] [warn]  +-->   User required --sinceForced='2020-04-14-09:55:32' but there is no match in source dataset nvpool/ROOT/hipster_2019.10-20200127T101549Z at /usr/share/src/znapzend/bin/../lib/ZnapZend.pm line 487.
...
````
note that although `die()` is used in the second case, this block of code is handled inside an `eval{}` and so throwing the error does not actually kill the whole script, just skips this dataset and the final cleanup. Not sure why it adds extra newlines after printing the message though.

**Comment by jimklimov** [comment]
(WIP for some other cases)

**Comment by oetiker** [comment]
please turn it into a PR when you are ready :)

**Comment by jimklimov** [comment]
Tested that `--since=X` mode does not destroy snapshots to ensure that "X" appears on destination, if there is already a newer common snapshot to continue resync from, but does ensure that "X" appears in destination history if destination's last common snapshot is older than "X" or destination was just autoCreated. It can also remove newer than "X" automatic snapshots if they are not (anymore) common, e.g. source was automatically cleaned by policy, and that way the "X" can also get put into destination history. 

Tested that `--sinceForced=X` mode can destroy snapshots on destination to ensure that "X" appears there, and can use a previous older common snapshot to increment from it (not from scratch) if there is one; can also resync from scratch otherwise. Well, that's what the user explicitly asked for in such case.

File: bin/znapzend [hunk 1]
```diff
@@ -71,6 +72,39 @@ sub main {
...
+        $opts->{forbidDestRollback} = 0;
+    }
+
+    if (defined($opts->{sinceForced}) && ($opts->{sinceForced} eq '')) {
+        delete $opts->{sinceForced};
+    }
+
+    if (defined($opts->{since}) && ($opts->{since} eq '')) {
+        delete $opts->{since};
+    }
```
oetiker, line 87: is this actually a problem ?

lib/ZnapZend.pm
```diff
@@ -366,8 +375,251 @@ my $sendRecvCleanup = sub {
...
+                                if (scalar @$srcSnapshots) {
+                                    my $dstSnapshots = $self->zZfs->listSnapshots($dstDataSet, $snapSendFilter);
+
+                                    my ($i, $snapName, $seenX, $seenD, $lastCommon, $lastCommonNum, $firstCommon, $firstCommonNum);
+                                    $lastCommon = undef; # track the newest common snapshot, if any
+                                    $lastCommonNum = undef; # Flips to "i" if we had any common snapshots
+                                    $firstCommon = undef; # track the oldest known common snapshot (if we can roll back to use it)
+                                    $firstCommonNum = undef; # Flips to "i" if we had any common snapshots
+                                    $seenX = undef; # Flips to "i" if we saw "X" before (or as) the newest common snapshot, looking from newest snapshots in src
+                                    $seenD = undef; # Flips to "i" if we saw "X" in DST during this search
```
oetiker, line 425: newly defined variables ARE undef by default 

describing what each one does seems cool, but why not just write

```perl
my $var; # description
```

lib/ZnapZend/ZFS.pm
```diff
@@ -424,9 +452,27 @@ sub sendRecvSnapshots {
...
+    }
+
+    # In certain cases, callers can set this argument to explicitly
+    # forbid (0), allow (1) or enforce if needed (2) a rollback of dest.
+    my $allowDestRollback = shift // undef;
+    if (!defined($allowDestRollback)) { $allowDestRollback = (!$self->sendRaw && !$self->forbidDestRollback) ; }
+
     my @recvOpt = $self->recvu ? qw(-u) : ();
-    push @recvOpt, '-F' if !$self->sendRaw;
+    push @recvOpt, '-F' if $allowDestRollback;
```
oetiker, line 475: is allowDestRollback connected to sendRaw ?

bin/znapzend
jimklimov: Did not check really, just for clarity in other code - if we have an input and go this way, the input makes sense at least for basics. Especially as we set this value from two options, so someone might (contrived) write `--since=X --sinceForced` thinking it is an option like that (probably getopt would protect from this one but still).

lib/ZnapZend.pm
jimklimov: I am uneasy about relying on defaults I guess, especially as I switch languages and compiler versions and nuances add bugs where not expected, and here with znapzend we expect to work on a decade or two of builds pf perl releases and I don't have them all to test thoroughly. Also in the code below I actively rely on them being `undef` so even if later defaults change, code can still rely in such explicit spelling.

Remember a time where a program misbehaved, but debugging did not expose anything; ended up that a C variable was not initialized so had garbage from RAM, but the debug-mode builds zeroed everything "helpfully".

lib/ZnapZend/ZFS.pm
jimklimov: I believe this partially originates from https://github.com/oetiker/znapzend/pull/496/files#diff-33cf5e498f84195d4308abccb7c59d70L355 adding the optionality of `-F` in relation to raw send (for encrypted datasets) and on another hand my PR adding `--forbidDestRollback` for people who want to explicitly not rollback existing snapshots for experiments or whatever reason.

Ultimately this PR makes it explicitly an optional argument of the `sendRecvSnapshot()` which can be set for each one sync operation in some cases (e.g. for intermediate desired snapshot in `--since` and `--sinceForced` modes) or if it is not set - defaults to the indications from global config options as they were directly used before.

lib/ZnapZend.pm
oetiker: :) This is not something new ... it has always been like that ... I am aware of the problem I tend to write unnecessary code as well just because I am not quite sure how it will work out in the current language ... but this one just looks very un-perlish or rather like something people would probably have done in perl4 (like 20 years ago or so)

should I fix it for you ?

bin/znapzend
oetiker: if it comes from getopt, then you can only have an empty string it it is something with which accepts a value ... so for 'since' that makes sense, but then you should also complain I think ... for since forced it does not as it does not accept a value in the first place ... 

the reason I am after these kinds of issues is becuase the more of them are in the code the harder it becomes to see the logic of the code and these unnecessary safeguards obscure the view ... 

this is why I like openapi for REST interfaces ... then I can be sure the openapi implementation sanitizes and validates the input before I go at it and I do not have to write extra safeguard and cleaning code locally ... 

getopt works in a similar way

lib/ZnapZend/ZFS.pm
oetiker: yes, if I remember correctly -F is incompatibly with sendRaw ... with this change the behavior of sendRaw will suffer, or am I missing something ?

bin/znapzend
jimklimov: Actually `--sinceForced=X` should work (and was tested to) as an explicitly dangerous replacement to `--since`. Maybe more conflict-checks and complaints make sense here. Technically still one can pass `--arg=''` to get an empty value in, and we don't want to bother about these.

lib/ZnapZend/ZFS.pm
jimklimov: I hope not, but can't test. Just a few lines above this is defined as:

````
    # In certain cases, callers can set this argument to explicitly
    # forbid (0), allow (1) or enforce if needed (2) a rollback of dest.
    my $allowDestRollback = shift // undef;
    if (!defined($allowDestRollback)) { $allowDestRollback = (!$self->sendRaw && !$self->forbidDestRollback) ; }
````

so normally (if not forced via since-mode) this would allow destination rollback if we are not in sendRaw mode (as before) and if user did not pass a flag to forbidDestRollback.

File: lib/ZnapZend.pm (hunk 2 above)
Comment (jimklimov) [reply lib/ZnapZend.pm:425]: If you insist, then yes please :) I'm hopping between branches and dayjob like crazy by now ;)

File: lib/ZnapZend/ZFS.pm (hunk 3 above)
Comment (oetiker) [reply lib/ZnapZend/ZFS.pm:475]: ok

File: lib/ZnapZend.pm (hunk 2 above)
Comment (oetiker) [inline lib/ZnapZend.pm:425]:
```suggestion
                                    my $i;
                                    my $snapName;
                                    my $lastCommon; # track the newest common snapshot, if any
                                    my $lastCommonNum; # Flips to "i" if we had any common snapshots
                                    my $firstCommon; # track the oldest known common snapshot (if we can roll back to use it)
                                    my $firstCommonNum ; # Flips to "i" if we had any common snapshots
                                    my $seenX; # Flips to "i" if we saw "X" before (or as) the newest common snapshot, looking from newest snapshots in src
                                    my $seenD; # Flips to "i" if we saw "X" in DST during this search
```

bin/znapzend
oetiker: ok

lib/ZnapZend.pm
jimklimov: Applied your patch... reluctantly :)

PR #498, opened 2020-07-29
Thirdparty bin/cpanm download issue

jimklimov:
I had some strange issues bootstrapping a build today, with wget failing to download `thirdparty/bin/cpanm` (curl did it okay reliably, wget succeeded once I think - all other times had a "SSL handshake failed"). Worked okay yesterday, so maybe the side is reconfiguring something (not all DNS aliases even responded).

The bigger-scale problem was that the original recipe left an empty bin/cpanm file which sufficed for subsequent `make` attempts to try using it and reported not having a `carton` to run - which was the issue I ended up hunting in vain.

This PR makes sure that either a not-empty `bin/cpanm` appears without errors from downloader, or it is removed and does not confuse `make` re-runs.

The rest of bootstrap went well with wget, after I implanted the cpanm fetched by curl.


oetiker:
hmm tests seem unhappy

jimklimov:
Typos. I hate typos. These are not the errors we are looking for...

jimklimov:
So... with my wget still not liking https://cpanmin.us today, I was left with no option but to let users `./configure` their `URL_CAT` if needed - either to prefer curl or wget (checking if available) with our presets, or providing their custom fetcher, or falling back to what we had to prefer wget first curl second whichever is there.

Issue #500, opened 2020-07-29
Selftests fail with "make check" but succeed with "./test.sh" and direct executions of "t/*.t" programs


jimklimov:
Console log of `make check` ends with:

````
Test Summary Report
-------------------
t/znapzend.t     (Wstat: 0 Tests: 51 Failed: 2)
  Failed tests:  49, 49
  Parse errors: Tests out of sequence.  Found (48) but expected (49)
                Plan (1..48) must be at the beginning or end of the TAP output
                Tests out of sequence.  Found (49) but expected (50)
                More than one plan found in TAP output
                Tests out of sequence.  Found (49) but expected (51)
Displayed the first 5 of 7 TAP syntax errors.
Re-run prove with the -p option to see them all.
t/znapzendzetup.t (Wstat: 0 Tests: 62 Failed: 0)
  Parse errors: Tests out of sequence.  Found (7) but expected (4)
                Tests out of sequence.  Found (8) but expected (5)
                Tests out of sequence.  Found (9) but expected (6)
                Tests out of sequence.  Found (10) but expected (7)
                Tests out of sequence.  Found (11) but expected (8)
Displayed the first 5 of 60 TAP syntax errors.
Re-run prove with the -p option to see them all.
Files=4, Tests=126, 344 wallclock secs ( 0.40 usr  0.35 sys + 128.90 cusr 95.65 csys = 225.30 CPU)
Result: FAIL
Failed 2/4 test programs. 2/126 subtests failed.
make: *** [Makefile:956: check] Error 255
````

Nothing apparent pops up in the megabyte of logs as to what failed, so looking deeper...

jimklimov:
A few suspicious gems from the pile of rubble:

````
t/znapzend.t .......                                                                                                                                                       
Failed -1/48 subtests                                                                                                                                                      
Possible attempt to separate words with commas at t/znapzendzetup.t line 143.                                                                                              
````

````
failed to destroy mock dataset(s)                                                                                                                                          
````

````
failed to list mock dataset(s)                                                                                                                                             
````

````
t/znapzendzetup.t ..                                                                                                                                                       
Failed 3/65 subtests                                                                                                                                                       
t/znapzendztatz.t .. ok                                                                                                                                                    
````

jimklimov:
* `make check` runs:
````
[...]

PR #501, opened 2020-07-29
Clean up tests and "make check" (see issue #500)

jimklimov:
At least a few inconsistencies are easy to spot in logs, let's start fixing #500 from these...


jimklimov:
As detailed in #500, one known issue that remains for `make check` at this point is the `test_harness` log behavior for daemonization tests (prints results twice).

Not sure why this did not trip us before, maybe the downloaded third-party ecosystem behaves differently after some years of evolution. At the moment I do not have good ideas how to address this bit well, so it is up for grabbing to other contributors :)

oetiker:
thanks

PR #502, opened 2020-07-30
Tests cleanup for daemonized mode

jimklimov:
Trying a follow-up for PR #501 addressing the #500 issue with `make check`, where `Test::More` returns two test results after checking daemonized mode and becomes confused.

Useful incantations for test-log inspection and comparison:
````
:; time make check 2>&1 | tee -a test-`date -u +%s`-check.log
````

````
:; time ./test.sh 2>&1 | tee -a test-`date -u +%s`-test.log
````

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -72,4 +72,10 @@ check:
 clean-selftest-rewritten:
 	$(RM) */*.selftest-rewritten */*/*.selftest-rewritten
 
-clean-local: clean-selftest-rewritten
+clean-pidfiles:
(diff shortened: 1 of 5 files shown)


jimklimov:
Huh, still `make check` failed for me on at least one system... could viability of this forked mess depend on versions of Test::something modules?..

At the very least, even if tests fail - these messages are now concentrated to `znapzend-daemonized.t` script which is sort of commented to be a fishy use-case for this harness.

PR #506, opened 2020-08-03
Modernize #384 : More flexible handling of offline destinations

jimklimov:
This picks up commits proposed in #384 and applies them to current master branch and its logical/conceptual changes.

The code passed locally a `./test.sh` run and from the looks of it - should be reasonably safe for common use-cases, but there may be caveats for non-default runs with `--inherited` mode which I suspect can end up ignoring this protection at the moment (detailed in my comments to the original PR).


oetiker:
cool

Harvie:
I am running znapzend 0.21.1 with "--cleanOffline" and it does not seem to delete the local snapshots as expected when destination is offline.

PR #507, opened 2020-08-03
Implement getSnapshotProperties() "inherit from same-named snapshot" concept

jimklimov:
This follows up from #384 and #506 to hopefully avoid a potential issue with "inherit" mode ignoring the parent's `dst_X_synced` values. 

Testing privately, review welcome to clear for merging. I hope it won't eat our data :)

### DIFF
--- a/lib/ZnapZend.pm
+++ b/lib/ZnapZend.pm
@@ -387,6 +387,11 @@ my $sendRecvCleanup = sub {
         # do not destroy data sets on the destination, or run post-send-command, unless all operations have been successful
         next if ($thisSendFailed);
 
+        # Remember which snapnames we already decided about in first phase
+        # (recursive cleanup from top backupSet-dst) if we did run it indeed.
+        # Do so with hash array for faster lookups into snapname existance.
+        my %snapnamesRecursive = ();
+
         #cleanup current destination
         if ($backupSet->{recursive} eq 'on') {
             # First we try to recursively (and atomically quickly)
@@ -399,6 +404,14 @@ my $sendRecvCleanup = sub {
             @snapshots = @{$self->zZfs->listSnapshots($backupSet->{$dst}, $backupSet->{snapCleanFilter})};
             $toDestroy = $self->zTime->getSnapshotsToDestroy(\@snapshots,
                          $backupSet->{"dst$key" . 'PlanHash'}, $backupSet->{tsformat}, $timeStamp, $self->since);
+
+            # Save the names we have seen, to not revisit them below for children
+            for (@{$self->zZfs->extractSnapshotNames(@snapshots)}) {
+                $snapnamesRecursive{$_} = 1;
+            }
+            for (@{$self->zZfs->extractSnapshotNames(@{$toDestroy})}) {
+                $snapnamesRecursive{$_} = 2;
+            }
 
             $self->zLog->debug('cleaning up snapshots recursively under destination ' . $backupSet->{$dst});
             {
@@ -430,6 +443,17 @@ my $sendRecvCleanup = sub {
             @snapshots = @{$self->zZfs->listSnapshots($dstDataSet, $backupSet->{snapCleanFilter})};
             $toDestroy = $self->zTime->getSnapshotsToDestroy(\@snapshots,
                          $backupSet->{"dst$key" . 'PlanHash'}, $backupSet->{tsformat}, $timeStamp, $self->since);
+
+            if (scalar(%snapnamesRecursive) > 0) {
+                for my $snapname (@{$self->zZfs->extractSnapshotNames(@{$toDestroy})}) {
+                    if ($snapnamesRecursive{$snapname}) {
+                        $self->zLog->debug('not considering whether to clean destination ' . $dstDataSet . '@' . $snapname . ' as it was already processed in recursive mode') if $self->debug;
+                        #print STDERR "DESTINATION CHILD UNCONSIDER CLEAN: BEFORE: " . Dumper($toDestroy) if $self->debug;
+                        @{$toDestroy} = grep { $snapname ne $_ } @{$toDestroy};
+                        #print STDERR "DESTINATION CHILD UNCONSIDER CLEAN: BEFORE: " . Dumper($toDestroy) if $self->debug;
+                    }
+                }
+            }
 
             next if (scalar (@{$toDestroy}) == 0);
 
@@ -475,7 +499,7 @@ my $sendRecvCleanup = sub {
         # cleanup source according to backup schedule
 
         # Remember which snapnames we already decided about in first phase
-        # (recursive cleanup from top backupSet) if we did run it indeed.
+        # (recursive cleanup from top backupSet-src) if we did run it indeed.
         # Do so with hash array for faster lookups into snapname existance.
         my %snapnamesRecursive = ();
 
--- a/lib/ZnapZend.pm
+++ b/lib/ZnapZend.pm
@@ -466,6 +466,10 @@ my $sendRecvCleanup = sub {
     else {
         # If "not sendFailed" or "cleanOffline requested"...
         # cleanup source according to backup schedule
+
+        # Remember which snapnames we already decided about in first phase
+        my %snapnamesRecursive = ();
+
         if ($backupSet->{recursive} eq 'on') {
             # First we try to recursively (and atomically quickly)
             # remove snapshots of "root" dataset with the recursive
@@ -479,6 +483,16 @@ my $sendRecvCleanup = sub {
             @snapshots = @{$self->zZfs->listSnapshots($backupSet->{src}, $backupSet->{snapCleanFilter})};
             $toDestroy = $self->zTime->getSnapshotsToDestroy(\@snapshots,
                          $backupSet->{srcPlanHash}, $backupSet->{tsformat}, $timeStamp);
+
+            # Save the names we have seen, to not revisit them below for children
+            foreach my $snapname (@snapshots) {
+                $snapname =~ s/^.*\@//;
+                $snapnamesRecursive{$snapname} = 1;
+            }
+            foreach my $snapname (@{$toDestroy}) {
+                $snapname =~ s/^.*\@//;
+                $snapnamesRecursive{$snapname} = 2;
+            }
 
             # preserve most recent common snapshots for each destination
             # (including offline destinations for which last-known sync
@@ -545,6 +559,14 @@ my $sendRecvCleanup = sub {
             @snapshots = @{$self->zZfs->listSnapshots($srcDataSet, $backupSet->{snapCleanFilter})};
             $toDestroy = $self->zTime->getSnapshotsToDestroy(\@snapshots,
                          $backupSet->{srcPlanHash}, $backupSet->{tsformat}, $timeStamp);
+
+            foreach my $snapname (@{$toDestroy}) {
+                $snapname =~ s/^.*\@//;
+                if ($snapnamesRecursive{$snapname}) {
+                    $self->zLog->debug('not considering source ' . $srcDataSet . '@' . $snapname . ' as it was already processed in recursive mode');
+                    @{$toDestroy} = grep { $snapname ne $_ } @{$toDestroy};
+                }
+            }
 
             # preserve most recent common snapshots for each destination
             # (including offline destinations for which last-known sync
--- a/lib/ZnapZend.pm
+++ b/lib/ZnapZend.pm
@@ -409,13 +409,19 @@ my $sendRecvCleanup = sub {
             for (@{$self->zZfs->extractSnapshotNames(@snapshots)}) {
                 $snapnamesRecursive{$_} = 1;
             }
-            for (@{$self->zZfs->extractSnapshotNames(@{$toDestroy})}) {
-                $snapnamesRecursive{$_} = 2;
-            }
+            # Note to devs: move extractSnapshotNames(toDestroy) up here
+            # if you would introduce logic that cleans that array.
 
             if (scalar($toDestroy) == 0) {
                 $self->zLog->debug('got an empty toDestroy list for cleaning up destination snapshots recursively under ' . $backupSet->{dst});
             } else {
+                # Note to devs: Unlike code for sources, here we only extract
+                # snapnames when we know the @toDestroy array is not empty -
+                # and it was not cleaned by any (missing) logic above.
+                for (@{$self->zZfs->extractSnapshotNames(@{$toDestroy})}) {
+                    $snapnamesRecursive{$_} = 2;
+                }
+
                 $self->zLog->debug('cleaning up snapshots recursively under destination ' . $backupSet->{$dst});
                 {
                     local $@;
--- a/lib/ZnapZend.pm
+++ b/lib/ZnapZend.pm
@@ -547,7 +547,7 @@ my $sendRecvCleanup = sub {
                 for my $dst (sort grep { /^dst_[^_]+$/ } keys %$backupSet){
                     my $dstDataSet = $backupSet->{src};
                     $dstDataSet =~ s/^\Q$backupSet->{src}\E/$backupSet->{$dst}/;
-                    my $recentCommon = $self->zZfs->mostRecentCommonSnapshot($backupSet->{src}, $dstDataSet, $dst, $backupSet->{snapCleanFilter}, ($backupSet->{recursive} eq 'on') );
+                    my $recentCommon = $self->zZfs->mostRecentCommonSnapshot($backupSet->{src}, $dstDataSet, $dst, $backupSet->{snapCleanFilter}, ($backupSet->{recursive} eq 'on'), undef );
                     if ($recentCommon) {
                         $self->zLog->debug('not cleaning up source ' . $recentCommon . ' recursively because it is needed by ' . $dstDataSet) if $self->debug;
                         #print STDERR "SOURCE RECURSIVE CLEAN: BEFORE: " . Dumper($toDestroy) if $self->debug;
@@ -628,7 +628,7 @@ my $sendRecvCleanup = sub {
             for my $dst (sort grep { /^dst_[^_]+$/ } keys %$backupSet){
                 my $dstDataSet = $srcDataSet;
                 $dstDataSet =~ s/^\Q$backupSet->{src}\E/$backupSet->{$dst}/;
-                my $recentCommon = $self->zZfs->mostRecentCommonSnapshot($srcDataSet, $dstDataSet, $dst, $backupSet->{snapCleanFilter});
+                my $recentCommon = $self->zZfs->mostRecentCommonSnapshot($srcDataSet, $dstDataSet, $dst, $backupSet->{snapCleanFilter}, undef);
                 if ($recentCommon) {
                     $self->zLog->debug('not cleaning up source ' . $recentCommon . ' because it is needed by ' . $dstDataSet) if $self->debug;
                     #print STDERR "SOURCE CHILD CLEAN: BEFORE: " . Dumper($toDestroy) if $self->debug;
--- a/lib/ZnapZend.pm
+++ b/lib/ZnapZend.pm
@@ -406,7 +406,7 @@ my $sendRecvCleanup = sub {
                          $backupSet->{"dst$key" . 'PlanHash'}, $backupSet->{tsformat}, $timeStamp, $self->since);
 
             # Save the names we have seen, to not revisit them below for children
-            for (@{$self->zZfs->extractSnapshotNames(@snapshots)}) {
+            for (@{$self->zZfs->extractSnapshotNames(\@snapshots)}) {
                 $snapnamesRecursive{$_} = 1;
             }
             # Note to devs: move extractSnapshotNames(toDestroy) up here
@@ -418,7 +418,7 @@ my $sendRecvCleanup = sub {
                 # Note to devs: Unlike code for sources, here we only extract
                 # snapnames when we know the @toDestroy array is not empty -
                 # and it was not cleaned by any (missing) logic above.
-                for (@{$self->zZfs->extractSnapshotNames(@{$toDestroy})}) {
+                for (@{$self->zZfs->extractSnapshotNames(\@{$toDestroy})}) {
                     $snapnamesRecursive{$_} = 2;
                 }
 
@@ -455,7 +455,7 @@ my $sendRecvCleanup = sub {
                          $backupSet->{"dst$key" . 'PlanHash'}, $backupSet->{tsformat}, $timeStamp, $self->since);
 
             if (scalar(%snapnamesRecursive) > 0) {
-                for my $snapname (@{$self->zZfs->extractSnapshotNames(@{$toDestroy})}) {
+                for my $snapname (@{$self->zZfs->extractSnapshotNames(\@{$toDestroy})}) {
                     if ($snapnamesRecursive{$snapname}) {
                         $self->zLog->debug('not considering whether to clean destination ' . $dstDataSet . '@' . $snapname . ' as it was already processed in recursive mode') if $self->debug;
                         #print STDERR "DESTINATION CHILD UNCONSIDER CLEAN: BEFORE: " . Dumper($toDestroy) if $self->debug;
@@ -528,10 +528,10 @@ my $sendRecvCleanup = sub {
                          $backupSet->{srcPlanHash}, $backupSet->{tsformat}, $timeStamp, $self->since);
 
             # Save the names we have seen, to not revisit them below for children
-            for (@{$self->zZfs->extractSnapshotNames(@snapshots)}) {
+            for (@{$self->zZfs->extractSnapshotNames(\@snapshots)}) {
                 $snapnamesRecursive{$_} = 1;
             }
-            for (@{$self->zZfs->extractSnapshotNames(@{$toDestroy})}) {
+            for (@{$self->zZfs->extractSnapshotNames(\@{$toDestroy})}) {
                 $snapnamesRecursive{$_} = 2;
             }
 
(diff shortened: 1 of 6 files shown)


jimklimov:
From testing:
* znapzending a dataset `nvpool/SHARED` which is a policy root:
````
[Mon Aug  3 17:12:55 2020] [debug] sending snapshots from nvpool/SHARED/opt/glassfish-4.0 to backup-adata/snapshots/nvpool/SHARED/opt/glassfish-4.0
# zfs list -H -o name -t snapshot -s creation -d 1 nvpool/SHARED/opt/glassfish-4.0
# zfs list -H -o name -t snapshot -s creation -d 1 backup-adata/snapshots/nvpool/SHARED/opt/glassfish-4.0
# zfs send -Lce -i 'nvpool/SHARED/opt/glassfish-4.0@znapzend-auto-2020-08-03T14:56:40Z' 'nvpool/SHARED/opt/glassfish-4.0@znapzend-auto-2020-08-03T15:08:18Z'|/usr/bin/amd64/mbuffer -q -s 256k -W 600 -m 1G|zfs recv -u -F 'backup-adata/snapshots/nvpool/SHARED/opt/glassfish-4.0'
# zfs list -H -o name -t snapshot nvpool/SHARED/opt/glassfish-4.0@znapzend-auto-2020-08-03T15:08:18Z
# zfs set org.znapzend:dst_0_synced=1 nvpool/SHARED/opt/glassfish-4.0@znapzend-auto-2020-08-03T15:08:18Z
# zfs set org.znapzend:dst_0=backup-adata/snapshots/nvpool/SHARED/opt/glassfish-4.0 nvpool/SHARED/opt/glassfish-4.0@znapzend-auto-2020-08-03T15:08:18Z
...
[Mon Aug  3 17:17:32 2020] [debug] got nothing to clean in children of source nvpool/SHARED
# zfs list -H -o name -t snapshot -s creation -d 1 nvpool/SHARED/opt/glassfish-4.0
# zfs list -H -o name -t snapshot -s creation -d 1 nvpool/SHARED/opt/glassfish-4.0
# zfs list -H -o name -t snapshot -s creation -d 1 backup-adata/snapshots/nvpool/SHARED/opt/glassfish-4.0
[Mon Aug  3 17:17:34 2020] [debug] not cleaning up source nvpool/SHARED/opt/glassfish-4.0@znapzend-auto-2020-08-03T15:08:18Z because it is needed by backup-adata/snapshots/nvpool/SHARED/opt/glassfish-4.0
[Mon Aug  3 17:17:34 2020] [debug] got nothing to clean in children of source nvpool/SHARED
...
````
From the above I see that it:
** successfully sets `dst_X_synced=1` property,
** honors this property by not cleaning the just-created snapshot,
** ...which it should not have cleaned anyway, so I suppose it should have gone looking to keep the older second-newest snapshot? or at least not inspected/logged this one?
** logs many times (for each child and their children) the verbatim same mantra "got nothing to clean in children of source nvpool/SHARED" (TODO: check the loop and message) UPDATE: fixed

**Comment by jimklimov** [comment]
For an inherited mode:

````
[Mon Aug  3 19:41:44 2020] [debug] now will look if there is anything to clean in children of destination backup-adata/snapshots/nvpool/SHARED/var/adm
# zfs list -H -o name -t snapshot -s creation -d 1 nvpool/SHARED/var/adm
# zfs list -H -o name -t snapshot -s creation -d 1 nvpool/SHARED/var/adm
# zfs get -H -s local,inherited -t snapshot -r -o property,value all nvpool/SHARED/var/adm@znapzend-auto-2020-08-03T17:41:40Z

### Here it is recursing into parent (hm, should somehow make snapshotExists quieter in case of missing snapshot):

# zfs list -H -o name -t snapshot nvpool/SHARED/var@znapzend-auto-2020-08-03T17:41:40Z
cannot open 'nvpool/SHARED/var@znapzend-auto-2020-08-03T17:41:40Z': dataset does not exist

[Mon Aug  3 19:41:45 2020] [debug] not cleaning up source nvpool/SHARED/var/adm@znapzend-auto-2020-08-03T17:41:40Z recursively because it is needed by backup-adata/snapshots/nvpool/SHARED/var/adm
[Mon Aug  3 19:41:45 2020] [debug] cleaning up source snapshots recursively under nvpool/SHARED/var/adm
# WOULD # zfs destroy -r nvpool/SHARED/var/adm@znapzend-auto-2020-08-03T16:37:20Z,znapzend-auto-2020-08-03T16:51:50Z,znapzend-auto-2020-08-03T16:58:52Z,znapzend-auto-2020-08-03T17:27:57Z,znapzend-auto-2020-08-03T17:33:11Z,znapzend-auto-2020-08-03T17:34:49Z
[Mon Aug  3 19:41:45 2020] [debug] now will look if there is anything to clean in children of source nvpool/SHARED/var/adm
[Mon Aug  3 19:41:45 2020] [info] done with backupset nvpool/SHARED/var/adm in 5 seconds
[Mon Aug  3 19:41:45 2020] [debug] send/receive worker for nvpool/SHARED/var/adm done (17830)
````

So seems to work and try to protect the snapshots now. Usually however this fast-tracks to the older `lastAndCommonSnapshots()` when both source and destination are up, instead of looping to call the new `getSnapshotProperties()` in `mostRecentCommonSnapshot()` -- to test that, have to neuter the clause for lastAndCommonSnapshots (`if (0) {...}`).

In either case it seems to protect the newest snapshot made (which is reasonable since it is replicated after all) and can consider dropping older snapshots (e.g. if I want 1 per hour, a 3-minute old one from earlier test is not needed anymore).

I did not yet manage to get it to try and delete a snapshot made earlier in backupSet's (`nvpool/SHARED`) history, maybe its scheduled time did not come yet.

Sometimes it ends up with this error however, when looking at a missing snapshot -- probably gotta merge hashes differently:
````
# zfs list -H -o name -t snapshot nvpool@znapzend-auto-2020-08-03T17:45:53Z
cannot open 'nvpool@znapzend-auto-2020-08-03T17:45:53Z': dataset does not exist
Reference found where even-sized list expected at /export/home/jim/shared/znapzend/bin/../lib/ZnapZend/ZFS.pm line 1035.
Reference found where even-sized list expected at /export/home/jim/shared/znapzend/bin/../lib/ZnapZend/ZFS.pm line 1035.
````

And also better find a way to not spew "dataset does not exist" to stderr where we anticipate it may not exist.

**Comment by jimklimov** [comment]
Algorithmic flaws: when we do get into the mode of looking for older `dst_X_synced` property flags (dest is offline so the just-made snapshot is not the newest common known), we now iterate into parent datasets to inspect their same-named snapshots and newly check if they have the property defined (and seems it is not inheritable into same-named snapshots of child datasets).

The check can optionally be done first in recursive mode, so zfs can destroy all same-named datasets in a tree starting from current backupSet->src root, and then we follow up by inspecting snapshots that remain in child datasets to see if any of those should be cleared (e.g. had earlier run inherited mode and have obsolete snapshots that are not defined in the root backupSet).

* we should not do this iterative approach for `dst_X_synced` if the current dataset is already the real backupSet (directly has the policy defined, we are not using the "--inherited" mode), and so not waste some processing time in the default mode for most users
* we should not re-check in child datasets the remaining snapshot names we already inspected (and decided to keep alive) in their parent, our current backupSet (inherited or root)
* it seems that when we descent into child datasets and hit a branch where several sub-datasets have same snapshot names, e.g. there was an "inherited" mode run from that point in tree, the current iteration goes from each such child and grandchild up in the tree, calling `zfs get` to look for `dst_X_synced`, and so re-parses the same data over and over (mostly finding nothing). While this is pedantically correct, it can easily spiral to be too costly in I/O, forking, processing... so there should be some safe and reliable way that we can cache the results and quickly see that "from this dataset@snapname upwards we know this property is defined to X, or is definitely not-set" and only request each such data once. And somehow not eat all RAM (fall back to current mode if lowMemRecursion?)...
* we should avoid getting all snapshot/dataset properties when we know we are going to use one or two specific names and discard the rest - pass those names to `zfs get propname1,propname2,... dataset@snap` right away.

**Comment by jimklimov** [comment]
Seems we have a nice bonus for the `dst_X_synced` use-case: the code also stores `dst_X` as a property of the snapshot that was the current backupSet source (e.g. for inherited ones too), which allows us to cut iterations short when we hit the true parent of a snapname:

````
# zfs get all nvpool/SHARED/var/adm@znapzend-auto-2020-08-03T16:51:50Z | grep local
nvpool/SHARED/var/adm@znapzend-auto-2020-08-03T16:51:50Z  org.znapzend:dst_0          backup-adata/snapshots/nvpool/SHARED/var/adm                       local
nvpool/SHARED/var/adm@znapzend-auto-2020-08-03T16:51:50Z  org.znapzend:dst_0_synced   1                                                                  local

:; zfs get all nvpool/SHARED/var/adm | grep local
:; zfs get all nvpool/SHARED/var | grep local
:; zfs get all nvpool/SHARED | grep local
nvpool/SHARED  org.znapzend:dst_0_plan     1weeks=>30minutes,2months=>1days,4months=>1weeks,10years=>1months  local
nvpool/SHARED  org.znapzend:dst_0          backup-adata/snapshots/nvpool/SHARED                               local
nvpool/SHARED  org.znapzend:mbuffer_size   1G                                                                 local
nvpool/SHARED  org.znapzend:src_plan       1weeks=>30minutes,2months=>1days,4months=>1weeks,10years=>1months  local
nvpool/SHARED  org.znapzend:zend_delay     0                                                                  local
nvpool/SHARED  org.znapzend:tsformat       znapzend-auto-%Y-%m-%dT%H:%M:%SZ                                   local
nvpool/SHARED  org.znapzend:post_znap_cmd  off                                                                local
nvpool/SHARED  org.znapzend:mbuffer        /usr/bin/amd64/mbuffer                                             local
nvpool/SHARED  org.znapzend:pre_znap_cmd   off                                                                local
nvpool/SHARED  org.znapzend:recursive      on                                                                 local
nvpool/SHARED  org.znapzend:enabled        on                                                                 local

```

lib/ZnapZend/ZFS.pm
```diff
@@ -553,7 +557,75 @@ sub sendRecvSnapshots {
...
+    #   string with our name-space propertyPrefix (e.g. "org.znapzend:...")
+    #   prepended if needed;
+    # * undef for defaulting to 'all'
+    # Returns a string safe to pass into `zfs get`
+    my $self = shift;
+    my $propnames = shift;
+
+    my $propertyPrefix = $self->propertyPrefix;
+
+    if (defined($propnames) && $propnames) {
```
oetiker, line 575: this does the same

```suggestion
    if ($propnames) {
```

lib/ZnapZend/ZFS.pm
```diff
@@ -553,7 +557,75 @@ sub sendRecvSnapshots {
...
+        if (ref($propnames) eq 'ARRAY') {
+            if (scalar(@$propnames) > 0) {
+                my $propstring = '';
+                foreach my $propname (@$propnames) {
```
oetiker, line 579:
```suggestion
                for my $propname (@$propnames) {
```

foreach is rather oldstyle

lib/ZnapZend/ZFS.pm
```diff
@@ -553,7 +557,75 @@ sub sendRecvSnapshots {
...
+                    next if !defined($propname);
```
oetiker, line 580:
```suggestion
                    next unless $propname;
```
since $propmane is probably never 0 or an empty string this should be fine

lib/ZnapZend/ZFS.pm
```diff
@@ -553,7 +557,75 @@ sub sendRecvSnapshots {
...
+                    chomp $propname;
+
+                    if ($propname eq 'all') {
+                        warn "=== getSnapshotProperties(): got an 'all' propname in array, any filtering will be moot";
```
oetiker, line 584: using 'warn' can be problematic as it does not go to the log ...

lib/ZnapZend/ZFS.pm
```diff
@@ -948,45 +1023,167 @@ sub getSnapshotProperties {
...
+                ###     `use v5.10.1;` or newer for ~~ smartmatch operator
+                ### Still warns about experimental perl feature in 5.22...
+                #if ($] >= 5.010001) {
+                #    if (@wantedPropnames ~~ \%properties) {
+                #        $inherit = 0;
+                #    }
+                #} else {
+                    # Old ways...
+                    my $uniqHits = 0;
+                    foreach my $key (keys %properties) {
```
oetiker, line 1141:
```suggestion
                    for my $key (sort keys %properties) {
```
sort keys for reproducability and use for

lib/ZnapZend/ZFS.pm
```diff
@@ -948,45 +1023,167 @@ sub getSnapshotProperties {
...
+                        if (exists($seenPropnames{$key})) {
```
oetiker, line 1142:
```suggestion
                        if (defined $seenPropnames{$key}) {
```

lib/ZnapZend/ZFS.pm
jimklimov: Would a bad caller with numeric `0` be a problem?
We have an explicit check for `eq ''` below, would that suffice if we do not check `defined` here?

oetiker: $propname with "", "0", 0 is false .. question is if that is a problem

lib/ZnapZend/ZFS.pm
jimklimov: What good would be sorting here? We anyway check all key values, so I guess it would only waste time to sort them?

lib/ZnapZend/ZFS.pm
jimklimov: Note to self: I'll revise the use of `print STDERR` too - there was a lot of such copypasta :(

lib/ZnapZend/ZFS.pm
jimklimov: Is there a particular difference (performance etc?) And also `defined (stuff)` vs. `defined stuff` - are these equivalent?

lib/ZnapZend/ZFS.pm
jimklimov: Well, this is supposed to be a sanity-checker of sorts for whatever other authors would put into their arrays eventually, maybe generated ones. Array elements that are `undef` or `""` we skip(ed) explicitly. If evolution provides a zfs property named `0` or `org.znapzend:0` we would fail to get it with this change...

lib/ZnapZend/ZFS.pm
jimklimov: So the shorter form is true if:
* propnames is defined,
* AND propnames is not (empty string OR zero number (OR maybe string "0" too?))

In particular, if it is NOT defined, there is no dereferencing error to check for further values?

lib/ZnapZend/ZFS.pm
oetiker: With exists you only know that the key exists ... the value might still be undefined
defined with brackets is equivalent ... that is just a style thing

lib/ZnapZend/ZFS.pm
oetiker: ok

lib/ZnapZend/ZFS.pm
oetiker: true :) I often use sort just to ensure that two runs are exactly the same ... but I guess it would not matter in this instance

lib/ZnapZend/ZFS.pm
jimklimov: So in the end, not applying this one, now - consciously :)

lib/ZnapZend/ZFS.pm
jimklimov: thanks

lib/ZnapZend/ZFS.pm
jimklimov: Uff, that was an unexpected long adventure :)

jimklimov:
At the moment it looks pretty neat and functional. There may be some points of improvement as listed in detail above, primarily:
* reduce unneeded I/O and processing load from requests to `zfs` command getting data we had seen recently
** stop recursing into parents when we pass a border (root backupSet, initial backupSet this run is iterating from) even if we do not find a way to cache all seen results
* reduce noise from `snapshotExists()` probing a snapshot that does not exist and we anticipate it is not there - find way to hide its STDERR into /dev/null if we asked for `$quiet==1`


but even so, performance does not seem to take extreme hits on reasonably small amounts of datasets, and when destination is online, and when we do not "runonce inherit" which is the primary mode of operation - so these and other improvements can happen in other PRs and by other authors ;)

lib/ZnapZend/ZFS.pm
jimklimov: Question aside, the proposed fix is merged

lib/ZnapZend.pm
```diff
@@ -535,14 +556,32 @@ my $sendRecvCleanup = sub {
...
             next if ($backupSet->{recursive} eq 'on' && $srcDataSet eq $backupSet->{src});
 
+            $self->zLog->debug('checking to clean up snapshots of source '. $srcDataSet);
+
             @snapshots = @{$self->zZfs->listSnapshots($srcDataSet, $backupSet->{snapCleanFilter})};
             $toDestroy = $self->zTime->getSnapshotsToDestroy(\@snapshots,
                          $backupSet->{srcPlanHash}, $backupSet->{tsformat}, $timeStamp);
 
+            for (@{$toDestroy}) {
+                my $snapname = $_;
```
oetiker, line 575: is this to make sure that the content of the $toDestroy array does not get changed ?

lib/ZnapZend.pm
```diff
@@ -474,11 +479,24 @@ my $sendRecvCleanup = sub {
...
             # and are all committed at once.
+            $self->zLog->debug('checking to clean up snapshots recursively from source '. $backupSet->{src});
 
             @snapshots = @{$self->zZfs->listSnapshots($backupSet->{src}, $backupSet->{snapCleanFilter})};
             $toDestroy = $self->zTime->getSnapshotsToDestroy(\@snapshots,
                          $backupSet->{srcPlanHash}, $backupSet->{tsformat}, $timeStamp);
 
+            # Save the names we have seen, to not revisit them below for children
+            for (@snapshots) {
+                my $snapname = $_;
```
oetiker, line 490: is this to make sure that the content of the snapshots array does not get changed ?

lib/ZnapZend.pm
```diff
@@ -474,11 +479,24 @@ my $sendRecvCleanup = sub {
...
+                $snapname =~ s/^.*\@//;
+                $snapnamesRecursive{$snapname} = 1;
+            }
+            for (@{$toDestroy}) {
+                my $snapname = $_;
```
oetiker, line 495: is this to make sure that the toDestroy array does not get changed ?

lib/ZnapZend.pm
oetiker: maybe having a function that takes care of extracting the snapshot names, and returning an array would make sense here ?

lib/ZnapZend/Config.pm
```diff
@@ -12,7 +12,11 @@ has zfsGetType => sub { 0 };
 has noaction => sub { 0 };
 has rootExec => sub { q{} };
 has timeWarp => sub { undef };
-has zLog => sub { Mojo::Exception->throw('zLog must be specified at creation time!') };
+has zLog            => sub {
```
oetiker, line 15: spacing

lib/ZnapZend/ZFS.pm
```diff
@@ -948,45 +1038,169 @@ sub getSnapshotProperties {
...
     }
+    my $numProps = keys %properties;
+
+    if ($self->debug) {
+        if ($numProps > 0) {
+            $self->zLog->debug("=== getSnapshotProperties(): GOT $inhMode (#$inherit) properties of $snapshot : " .Dumper(\%properties) );
+        }
+    }
+
+    if ($inherit == 2 || $inherit == 3) {
```
oetiker, line 1110: I find such long multi level if/then constructs a bit hard to follow ... how about returning the `\%properties if $inherit < 2` here to loose a level ?

lib/ZnapZend.pm
jimklimov: good point, this bit of code happens a few times too many :)

lib/ZnapZend.pm
jimklimov: yes, I found that with `for my $snapname (@toDestroy)` at least, regexing the `$snapname` in the next line did change the array entries from `dataset@snapname` to `snapname`.

lib/ZnapZend/ZFS.pm
jimklimov: I originally thought to make it a bit mask, then settled for a number of magic values with specific meanings (commented above) that are a bit faster to compare than more meaningful strings or further variables - and we pass this codepath a lot when we do :\

oetiker: oh I did not realize that ... please don't :) use a hash with speaking  keys for the features you want enabled ...

lib/ZnapZend.pm
jimklimov: Although not sure on second thought: I tracked twice the snapshots we considered (in filtered `@snapshots` and selection `@toDestroy`) with different values... I guess there was an idea to differentiate the two though not sure now, probably it did not get into code and popped out of memory at the moment. So we'd need then to merge the two resulting arrays also, instead of now just writing into one. Or pass several to one routine...

lib/ZnapZend/ZFS.pm
jimklimov: I think I get the idea but did not find an example to rip off, do you have some handy please?

jimklimov:
Another TODO (now or in another PR to not stall this one): if the idea with not-revisiting snapshot cleanup for tags we have considered in recursive mode seems okay, it can be replicated into destination clean-ups too.

lib/ZnapZend/ZFS.pm
oetiker:
```perl
my $inherit = {
    all => 1,
    greens => 0,
    reds => 1
};


if ($inherit->{all} or $inherit->{greens}){ .... 

}
```

lib/ZnapZend.pm
oetiker: What I mean is that instead of walking the array and running the regexp to extract the snapshot names you could create a function which does that

```perl 
sub extractSnapshotNames($array){
      my @ret;
      for (@$array){
           /\@(.+)/ and push @ret, $1;
     }
     return \@ret;
}
```

lib/ZnapZend.pm
oetiker: this is a feature :) see my proposed function above

lib/ZnapZend/ZFS.pm
jimklimov: Ah, just so? I thought there was something computationally cheaper, like C enums. Thanks, will try.

lib/ZnapZend.pm
jimklimov: Thanks, I'll see if a usecase remains for tracking that as a hash with various levels, or an array is okay.

jimklimov: I guess the hash usage originated with another sanity check, to make sure values are unique just in case. And since we had a hash, I decided to keep it, e.g. to have speedy lookups into the keys (is this snapname in the list?)...

jimklimov: Done, pushed, tested

lib/ZnapZend/ZFS.pm
jimklimov: I suppose my fix is not what you wanted (ended up like an enum), but cursorily seems to work and steps away from arcane numbers to human names for them, but still uses the number underneath. Might revise later to have it as a flag map per your example, but I think it would be a refactor irrelevant to functionality at the moment :)

lib/ZnapZend/ZFS.pm
```diff
@@ -213,6 +227,46 @@ sub listSnapshots {
     return \@snapshots;
 }
 
+sub extractSnapshotNames {
+    # Routines like listSnapshots() operate with fully qualified
+    # "dataset@snapname" strings as returned by ZFS. For some data
+    # matching we need to compare just the "snapname" lists.
+    my $self = shift;
+    my ($array) = @_; # Note: call with extractSnapshotNames(\@arrVarName) !
```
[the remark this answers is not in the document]
jimklimov: I tried that originally, seems this was the syntax used in other routines, but seems it was about as messy (especially with need to pass `\@arrname` vs `@arrname`. 

Neither approach handled arguments coming from `qw(word list)` for example; I hoped this variant to catch cases where someone passes multiple single string arguments by mistake or somehow else, but must have got perl syntax wrong somewhere.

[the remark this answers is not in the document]
jimklimov: Re-checked to be sure, behaves the same in the three use-cases commented after the routine :)

jimklimov:
Hm, seems like we say the snapshots seen in recursive mode would not be revisited in toDestroy list, but then the list still contains them. Gotta check that this grep works...

jimklimov:
Looks better now, almost no snaps to clean in the child phase because they were looked at in the recursive phase and decidedly left to stick around.

jimklimov:
Interesting, I think I over-engineered my worries for now: in manual experiments I was using `zfs send -R -I ... | zfs recv` for quicker wholesale replications -- something I intended or asked long time ago to use in `znapzend` for first stage sendings (like recursive clean-ups, that might not need following by child-by-child cleanups if all went well), so I assumed somehow this was the mode it used already - and then it would only mark the topmost parent dataset of a replication run with `dst_X_synced` flags in `sendRecvSnapshot()`.

Well, we're saved by the bell: that mode is not in codebase yet, so every dataset in a subtree is replicated individually, marked as synced individually, and iteration to get the snapshot properties involved does not go far and cost much even without us making some caching and tree-pruning :)

jimklimov:
Applied a few patches from #497 that do fix issues and would conflict the remaining PR after merging one to master first.

lib/ZnapZend/ZFS.pm
```diff
@@ -12,12 +12,12 @@ use Data::Dumper;
...
 ### properties inherited from higher level datasets; but for
 ### snapshots it only returns the same - not from higher snaps.
-my %inheritLevels = {
+my %inheritLevels = (
     local_only => 0,
     local_zfsinherit => 1,
     local_recurseparent => 2,
     local_recurseparent_zfsinherit => 3,
-};
+);
```
oetiker, line 20: as you mentioned :) sort of semi satisfying this way ...

jimklimov: This PR is stalling and growing too large to review well already :( Can we leave such refinements for later iterations? I believe after the "proper fix" functionality would be the same, maintainability (key words instead of magic numbers) too, so that classifies as a refactor and can be separated :)

oetiker: Sure if you would like to follow up with another one to fix this properly ... I am fine with merging this one now. 

The reason I am trying to catch these things is that they are non-ideomatic and tend to make my life harder when I have to figure out some odd behavior some time down the road. As I tend to be the one stuck with these projects in the long run.

oetiker:
thanks for your contribution

Issue #479, opened 2020-02-28
Global symbol "$dataSet" requires explicit package name


c0deaddict:
When I'm trying to run the `master` branch on NixOS I get this warning:

```
Global symbol "$dataSet" requires explicit package name (did you forget to declare "my $dataSet"?) at /nix/store/rlvxvxa9xayjz623g72awiq96j54ii11-znapzend-0.20.0alpha/lib/ZnapZend/Config.pm line 92.
Compilation failed in require at /nix/store/rlvxvxa9xayjz623g72awiq96j54ii11-znapzend-0.20.0alpha/lib/ZnapZend.pm line 6.
BEGIN failed--compilation aborted at /nix/store/rlvxvxa9xayjz623g72awiq96j54ii11-znapzend-0.20.0alpha/lib/ZnapZend.pm line 6.
Compilation failed in require at /nix/store/rlvxvxa9xayjz623g72awiq96j54ii11-znapzend-0.20.0alpha/bin/znapzend line 9.
BEGIN failed--compilation aborted at /nix/store/rlvxvxa9xayjz623g72awiq96j54ii11-znapzend-0.20.0alpha/bin/znapzend line 9.
```

Looks like this has been caused by this commit: https://github.com/oetiker/znapzend/commit/60e7afe76ed375c0b3994c29cd2283c02e0f59c4 The GitHub actions annotations are also not happy about this change.

I don't have any experience with Perl, otherwise I would make a PR.

oetiker:
fix is in pr #480

c0deaddict:
Nice! Thanks for the very quick response :)

jimklimov:
Better late than never: "sorry, my bad". Seems like a copypaste issue slipped through, should have been `$backupSet->{src}` instead of `$dataSet` that was relevant elsewhere :(

jimklimov:
Hopefully this is closeable now (or soon after vetting ;) )

c0deaddict:
Yes, problem is solved and this ticket can be closed :+1:

PR #508, opened 2020-08-07
Fix regression #479-bis

jimklimov:
Restores the failsafe removed quickly in #480, this time applying proper variable naming to the checked item.

### DIFF
--- a/lib/ZnapZend/Config.pm
+++ b/lib/ZnapZend/Config.pm
@@ -103,9 +103,9 @@ my $checkBackupSets = sub {
             # dataSetExists() below, so just ignore this entry.
             # If we really do get here, take a hard look at recursive
             # and/or inherited modes for run-once.
-            print STDERR "#checkBackupSets# SKIP backupSet='" . $backupSet->{src} .
-                "' because it is not a filesystem,volume. " .
-                "BUG: Should not get here.\n";# if $self->debug;
+            $self->zLog->error( "#checkBackupSets# SKIP backupSet='"
+                . $backupSet->{src} ."' because it is not a filesystem,volume. "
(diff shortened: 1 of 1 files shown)


lib/ZnapZend/Config.pm
```diff
@@ -80,15 +80,30 @@ my $checkBackupSets = sub {
...
         }
 
+        if ( $backupSet->{src} =~ m/[\@]/ ) {
+            # If we are here, somebody fed us a snapshot in the list of
+            # datasets, which is likely a bug elsewhere in discovery.
+            # We do not want to fail whole backup below due to faulted
+            # dataSetExists() below, so just ignore this entry.
+            # If we really do get here, take a hard look at recursive
+            # and/or inherited modes for run-once.
+            print STDERR "#checkBackupSets# SKIP backupSet='" . $backupSet->{src} .
```
oetiker, line 101: not using zLog here ?

jimklimov: I thought about it a little... this PR is essentially a `git revert` + typo fix, so I did not go too deep. Given that this catches a situation that should never happen at all, I am not sure what could be proper logging here.

oetiker: well the kind of logging that is best suited to catch the attendents attention ... zLog->error I would suggest ... after all they should be watching the znapzend logs

jimklimov: Done, checked that `$self->zLog->error()` by itself is not fatal to code progressing ;)

jimklimov:
Note: the check for snapshot in this block may be among the targets for #514 to check for bookmarks as another "unexpected" case... But that is a separate issue to investigate and tackle globally.

Issue #503, opened 2020-07-31
There seems to be a problem replicating datasets that are zfs-promoting branches part of bigger tree (rpool/ROOT, zones...)


jimklimov:
As I was checking how my backups went, I found that replicas of my rootfs related datasets are not branches of each other, but seem to be unique data histories (with no "origin" set in each of them on the backup side; I suppose `--autoCreation` was involved, my configs have it enabled by default). This causes confusion at least when a backup pool is not regularly accessible, and older automatic snapshots get deleted, and there is no common point to resync from (not even among snapshot made by beadm, etc.). 

This may also waste space on backup pool, writing many copies of rootfs instead of building on top of shared history, but I am not fully sure about that bit (my backup pool claims a high dedup ratio while dedup is off in dataset settings).

In practice this impacts primarily rpool/ROOT/something(/subsets) and pool/zones/zonerootnames/... that are automatically shuffled as a particular rootfs gets activated (so whatever is related to the current rootfs is promoted as the main tree of ZFS "live" data, and older roots are origined at its snapshots), but I suppose other zfs promotions are similarly susceptible.

My guess is that autocreation should check if the newly found dataset has an origin snapshot attribute, and issue a replication going from that point not from scratch. The "origin" key word is not seen in znapzend codebase ;)
* There are likely to be nuances - e.g. apparent riddle is what if the dataset pointed to by the origin field is not currently on the remote destination?
** Is it going to appear, e.g. as part of same autocreation run? => re-schedule it to appear first, at least up to this snapshot? (maybe recurse through further origins as needed)
** Does it have some other znapzend policy on source, so can appear from another run? => use that knowledge somehow? or is it a breach of whatever sanity in place? just do a `runonce --inherited` on that other policy, and wait for it to complete?
** Can it make sense to send the current (non-"original") dataset in a way similar to that explored for --sinceForced=... from #497 to make sure that even if it appears as an independent history of incremental snapshots, that history includes the origin snapshot (probably not named by znapzend timestamped patterns) so that if the user later adds that origin to a backed-up policy, it could be re-branched from such point. Common history is common, whoever owns it at the moment.
** Oh, and there are also subsequent zfs promotions shuffling around the current owners of historic snapshots and their consumers who branched off that point...
[...]

PR #512, opened 2020-08-12
Contrib PoC script for issue #503 and README updates

jimklimov:
The script is not production quality and does not even cover all the cases I've seen, but on one hand it allows to rectify some of the problems with simple single-dataset ZBEs already (running it now locally to fix my destination of zones' backups), and on another can serve as a basis for proper implementation inside znapzend itself.

The updated README.md and its markup can be seen rendered from markdown e.g. in my branch source for this PR at https://github.com/jimklimov/znapzend/tree/contrib-issue-503

### DIFF
--- a/README.md
+++ b/README.md
@@ -108,6 +108,10 @@ http://pkg.openindiana.org/hipster/en/search.shtml?token=znapzend&action=Search
 made with the recipe at
 https://github.com/OpenIndiana/oi-userland/tree/oi/hipster/components/sysutils/znapzend
 
+```sh
+pkg install backup/znapzend
+```
+
 Configuration
 -------------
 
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -40,41 +40,54 @@ On RedHat you get the necessaries with:
 
 On Ubuntu / Debian with:
 
-    apt-get install perl unzip
+    apt-get install perl unzip autoconf carton
```
oetiker, line 43: hmmm autoconf is only required when building the git checkout ... carton should never be required

jimklimov: Hmm... I always started with bootstrap (unlike the README instructions) so indeed autoconf may be an overshot. Apparently this (both) is also what @Dacesilian used on Gitter.

No opinion about carton, except that maybe using an OS-provided one is somehow better for support/compat/etc?

jimklimov: Curious, the release zip file does not include `configure` script, but the tarball does.
It is both tracked in git SCM and excluded in `.gitignore`, if that helps.

jimklimov: Amended the README to convey that bootstrap exists and then might need autoconf and family to be used.

Issue #516, opened 2020-08-17
Fix inheritLevels semantics from #507


jimklimov:
...and also this message from `perl -w` runs:

````
Variable "%inheritLevels" will not stay shared at .../znapzend/t/../lib/ZnapZend/ZFS.pm line 475.
Variable "%inheritLevels" will not stay shared at .../znapzend/t/../lib/ZnapZend/ZFS.pm line 1228.
````

jimklimov:
From discussions in #507 : 

> @jimklimov 13 days ago Author Contributor
> I originally thought to make it a bit mask, then settled for a number of magic values with specific meanings (commented above) that are a bit faster to compare than more meaningful strings or further variables - and we pass this codepath a lot when we do :\

> @oetiker 12 days ago Owner
> oh I did not realize that ... please don't :) use a hash with speaking keys for the features you want enabled ...

> @jimklimov I think I get the idea but did not find an example to rip off, do you have some handy please?

> @oetiker 
````
my $inherit = {
    all => 1,
    greens => 0,
    reds => 1
};


if ($inherit->{all} or $inherit->{greens}){ .... 
}
````

> @jimklimov
> Ah, just so? I thought there was something computationally cheaper, like C enums. Thanks, will try.
> UPDATE: Oh, these are not strings, so might be cheap...
> I suppose my ultimate fix is not what you wanted (ended up like an enum), but cursorily seems to work and steps away from arcane numbers to human names for them, but still uses the number underneath. Might revise later to have it as a flag map per your example, but I think it would be a refactor irrelevant to functionality at the moment :)

> @oetiker 8 days ago Owner
> as you mentioned :) sort of semi satisfying this way ...
> Sure if you would like to follow up with another one to fix this properly ... I am fine with merging this one now.
> The reason I am trying to catch these things is that they are non-ideomatic and tend to make my life harder when I have to figure out some odd behavior some time down the road. As I tend to be the one stuck with these projects in the long run.

jimklimov:
Should be fixed after merging #519

PR #517, opened 2020-08-17
Tests cleanup - really test daemonized mode behaviors (complete nits for #500)

jimklimov:
Follows up from #501 and #502 to complete the issues found with #500 : `make check` now passes and the daemonized behaviors (e.g. clash over a pidfile) are now inspected in parent and child forks.

### DIFF
--- a/lib/ZnapZend/ZFS.pm
+++ b/lib/ZnapZend/ZFS.pm
@@ -12,7 +12,7 @@ use Data::Dumper;
 ### For "usual" datasets (filesystem,volume) `zfs` returns the
 ### properties inherited from higher level datasets; but for
 ### snapshots it only returns the same - not from higher snaps.
-my %inheritLevels = (
+our %inheritLevels = (
     local_only => 0,
     local_zfsinherit => 1,
     local_recurseparent => 2,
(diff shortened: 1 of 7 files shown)


lib/ZnapZend/ZFS.pm
```diff
@@ -12,7 +12,7 @@ use Data::Dumper;
 ### For "usual" datasets (filesystem,volume) `zfs` returns the
 ### properties inherited from higher level datasets; but for
 ### snapshots it only returns the same - not from higher snaps.
-my %inheritLevels = (
+our %inheritLevels = (
```
oetiker, line 15: note that the numbers in my example #516 was that you could turn them on or off by writing one or zero as a value ...

jimklimov: I am still puzzled about proper (and type/name safe) passing of such "inherit" values as arguments to the function which can then be inspected. Looked hard at Class::struct to define and pass new() instances of that and checking whether this or that class member is true or false... would that be the way?

oetiker: with our you make this information accessible from the outside (%ZnapZend::ZFS::inheritLevels) ... note also that it is a Class variable and not an instance variable ... is this what you want?

jimklimov: Well, in the current use effectively as an enum of distinct values to compare numeric inherit arguments/variables against, I don't think the nuance matters much. At least until someone decides to change its values on the fly :)
I guess a separate struct is cleaner for such purpose, use and concept alike.

Issue #499, opened 2020-07-29
Optionally email the admin if a send task failed and we suspend the source cleanup


jimklimov:
A recurrent problem at one of sites I help manage is that due to whatever reasons, they occasionally lose connection to the NAS which hosts a backup pool, so on the server side this pool becomes read-only. Recovery varies on situation, but is something that an admin should do anyway - until then, regular snapshots of the original system accumulate on the storage until zfs performance collapses due to scarce free space fragmentation. One way out of this practical problem is to know about it in time, and untangle the problems with the destination pool before overrunning the source pool impacts the users. Of course there may be proper reactive monitoring setups and syslog diggers and stuff, but the actual subsystem reporting a problem (and in a way causing it) is also a good place to proactively inform the admins.

Earlier I extended the "cleanup source" error handling to (re-)list all problematic datasets in one place when we skip the cleanup (and not just state we had a problem) with the text ending up on console or in syslog, at https://github.com/oetiker/znapzend/blame/1caafdbb48b807ab8d4c2539f7987ea0477af9c5/lib/ZnapZend.pm#L458 - so for me the "obvious idea" is to add an optional CLI argument for admin e-mail, and if that is configured - try to use system `mailx` or some such CLI utility, and submit essentially the same text. Then it is up to local mail submission program and relaying setup (outside `znapzend` concerns and complexity) to deliver this somewhere if it should end up in a mailbox not on the same host. (in other words, I do not want to teach znapzend to bother about networked SMTP, AUTH, STARTTLS and whatever comes with e-mail these days).

Are there any apparent issues with such a plan?

jimklimov:
Saw #221 after posting this suggestion but after re-reading that PR, not sure it would be a duplicate :)

PR #221 seems to add ability to log *everything* emitted into a file or pipe, same as done for console (stderr/stdout) or syslog, and #499 concerns about alerting to fail to send and so refusing to clean up origin (which may lead to it getting too full).

jimklimov:
Goal achieved by PR above :)

PR #520, opened 2020-08-19
Optionally email the admin if a send task failed and we suspend the source cleanup

jimklimov:
Should send the error summary to specified recipient(s) and fix the issue #499 as far as I (original poster) am concerned :) 

Tested on a local system, delivers to one or two (comma-separated) local mailboxes using the standard `/usr/lib/sendmail` or `/usr/sbin/sendmail` alias (exim, postfix also provide that interface). Internet delivery is up to MTA setup and outside znapzend's interest.

````
:; znapzend --noaction --runonce=nvpool/SHARED/var/test --inherited --debug --cleanOffline --mailErrorSummaryTo=root,jim

...
[2020-08-20 00:29:08.80293] [26384] [warn] ERROR: 1 send task(s) below failed for nvpool/SHARED/var/test, but "cleanOffline" mode is on, so proceeding to clean up source dataset carefully:
[2020-08-20 00:29:08.80315] [26384] [warn]  +-->   destination 'backup-adata/snapshots/nvpool/SHARED/var/test' does not exist or is offline. ignoring it for this round...
[2020-08-20 00:29:08.80643] [26384] [warn] Sending a copy of the report above to root,jim
[2020-08-20 00:29:08.83905] [26384] [debug] checking to clean up snapshots recursively from source nvpool/SHARED/var/test
...
````

If the mail program does not exist in standard path (currently hardcoded, can be made an argument later), it is noted but not fatal:
````
[2020-08-20 00:27:28.54081] [26359] [warn] ERROR: 1 send task(s) below failed for nvpool/SHARED/var/test, but "cleanOffline" mode is on, so proceeding to clean up source dataset carefully:
[2020-08-20 00:27:28.54089] [26359] [warn]  +-->   destination 'backup-adata/snapshots/nvpool/SHARED/var/test' does not exist or is offline. ignoring it for this round...
Can't exec "/usr/sbin/sendmailsss": No such file or directory at /export/home/jim/shared/znapzend/bin/../lib/ZnapZend.pm line 786.
Can't open /usr/sbin/sendmailsss to send a copy of the report above to root,jim!
````

Message in the box:
````
From jim@jimoi.local Thu Aug 20 00:29:08 2020
Return-Path: <jim@jimoi.local>
Received: from jimoi.local (jimoi [127.0.0.1])
        by jimoi.local (8.15.2+Sun/8.15.2) with ESMTP id 07JMT880026388;
        Thu, 20 Aug 2020 00:29:08 +0200 (CEST)
Received: (from jim@localhost)
        by jimoi.local (8.15.2+Sun/8.15.2/Submit) id 07JMT8wN026387;
        Thu, 20 Aug 2020 00:29:08 +0200 (CEST)
Date: Thu, 20 Aug 2020 00:29:08 +0200 (CEST)
From: jim <jim@jimoi.local>
Message-Id: <202008192229.07JMT8wN026387@jimoi.local>
To: root@jimoi.local, jim@jimoi.local
Subject: znapzend replication error summary
Content-Length: 290

-------
ERROR: 1 send task(s) below failed for nvpool/SHARED/var/test, but "cleanOffline" mode is on, so proceeding to clean up source dataset carefully:
 +-->   destination 'backup-adata/snapshots/nvpool/SHARED/var/test' does not exist or is offline. ignoring it for this round...
-------
````

### DIFF
--- a/lib/ZnapZend.pm
+++ b/lib/ZnapZend.pm
@@ -752,14 +752,25 @@ my $sendRecvCleanup = sub {
     }
 
     #cleanup source
-    if (scalar(@sendFailed) > 0 and not $self->cleanOffline) {
-        $self->zLog->warn('ERROR: suspending cleanup source dataset because '
-            . scalar(@sendFailed) . ' send task(s) failed:');
-        for my $errmsg (@sendFailed) {
+    #we want the message summarizing errors regardless of continuing to cleanup
+    if (scalar(@sendFailed) > 0) {
+        my $errmsg;
+        if ($self->cleanOffline) {
+            # Note: this is about all transfer failures, such as offline dest
+            # or it is full, or read-only, or source too full to make a snap...
+            $errmsg = 'ERROR: ' . scalar(@sendFailed) . ' send task(s) below failed,' .
+                ' but "cleanOffline" mode is on so proceeding to cleanup source dataset carefully:' ;
+        } else {
+            $errmsg = 'ERROR: suspending cleanup source dataset because ' .
+                scalar(@sendFailed) . ' send task(s) failed:' ;
+        }
+        $self->zLog->warn($errmsg);
+        for $errmsg (@sendFailed) {
             $self->zLog->warn(' +-->   ' . $errmsg);
         }
     }
-    else {
+
+    if (scalar(@sendFailed) == 0 or $self->cleanOffline) {
         # If "not sendFailed" or "cleanOffline requested"...
         # cleanup source according to backup schedule
 
--- a/lib/ZnapZend.pm
+++ b/lib/ZnapZend.pm
@@ -53,7 +53,7 @@ has nodelay                 => sub { 0 };
 has skipOnPreSnapCmdFail    => sub { 0 };
 has skipOnPreSendCmdFail    => sub { 0 };
 has cleanOffline            => sub { 0 };
-has mailErrorSummaryTo      => sub { q{} };
+has 'mailErrorSummaryTo';
 has backupSets              => sub { [] };
 
 has zConfig => sub {
--- a/lib/ZnapZend.pm
+++ b/lib/ZnapZend.pm
@@ -769,7 +769,7 @@ my $sendRecvCleanup = sub {
                 scalar(@sendFailed) . ' send task(s) failed:' ;
         }
         $self->zLog->warn($errline);
-        if ($self->mailErrorSummaryTo ne '') {
+        if ($self->mailErrorSummaryTo) {
             $errmsg = $errline . "\n";
         }
         for $errline (@sendFailed) {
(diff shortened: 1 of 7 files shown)

### CODE REVIEW

File: lib/ZnapZend.pm [hunk 1]
```diff
@@ -53,6 +53,7 @@ has nodelay                 => sub { 0 };
 has skipOnPreSnapCmdFail    => sub { 0 };
 has skipOnPreSendCmdFail    => sub { 0 };
 has cleanOffline            => sub { 0 };
+has mailErrorSummaryTo      => sub { q{} };
```
oetiker, line 56:
```suggestion
has 'mailErrorSummaryTo';
```

lib/ZnapZend.pm
```diff
@@ -752,14 +753,53 @@ my $sendRecvCleanup = sub {
...
+            $errline = 'ERROR: ' . scalar(@sendFailed) . ' send task(s) below ' .
+                'failed for ' . $backupSet->{src} . ', but "cleanOffline" mode ' .
+                'is on, so proceeding to clean up source dataset carefully:' ;
+        } else {
+            $errline = 'ERROR: suspending cleanup source dataset ' .
+                $backupSet->{src} . ' because ' .
+                scalar(@sendFailed) . ' send task(s) failed:' ;
+        }
+        $self->zLog->warn($errline);
+        if ($self->mailErrorSummaryTo ne '') {
```
oetiker, line 772:
```suggestion
        if ($self->mailErrorSummaryTo) {
```

lib/ZnapZend.pm
```diff
@@ -752,14 +753,53 @@ my $sendRecvCleanup = sub {
...
+            $errmsg = $errline . "\n";
+        }
+        for $errline (@sendFailed) {
+            $self->zLog->warn(' +-->   ' . $errline);
+            if ($self->mailErrorSummaryTo ne '') {
+                $errmsg .= ' +-->   ' . $errline . "\n";
+            }
+        }
+
+        if ($self->mailErrorSummaryTo ne '') {
```
oetiker, line 782:
```suggestion
        if ($self->mailErrorSummaryTo) {
```

oetiker (changes requested):
Setting to an empty string and then comparing to it seems rather non-ideomatic ... and since a valid value will never be '0' there should be no problem with just using the value directly ...

lib/ZnapZend.pm
jimklimov: Good point; so far I was just following the precedent in nearby lines :)

oetiker (changes requested):

jimklimov:
Bump? This change is in our production for a month now :)

oetiker (approved):
oops missed that

PR #525, opened 2020-10-12
Fix typo and addes destroy options on recv side for running as unpriv…

GabrieleV:
…ileged user

Running as an unprivileged user needs the destroy permission on the remote receiving destination to allow cleaning expired snapshots.


oetiker (approved):

oetiker:
thanks

PR #528, opened 2020-10-20
Problem: Hard to use znapzend dev bins in a web of symlinks

jimklimov:
Solution: Follow up from 75857808ab5fc1030ac294c4e3634d4b5ef323c9
which added the line which is removed during "proper" installation,
but now use FindBin::RealBin instead of just ::Bin to allow developers
to shoot themselves in the foot easier by symlinking their version
in progress (or a snapshot of the development workspace) right into
their `/usr/bin/znapzend*` pathnames.

### DIFF
--- a/bin/znapzend
+++ b/bin/znapzend
@@ -1,7 +1,7 @@
 #!/usr/bin/env perl
 
 use lib qw(); # PERL5LIB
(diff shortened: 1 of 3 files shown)


oetiker (approved):
yes !!!

PR #552, opened 2021-06-30
contrib/synczbe: better support for "history owner" promotion

jimklimov:
Some time ago I started collecting notes and throwing together a bit of automation for dataset trees like rootfs and zoneroots, which are automatically cloned and promoted on the system side as updates are installed -- but their backups done with `znapzend` tend to "autoCreate" new datasets from scratch rather than take advantage of gigabytes already present (under a different name, that of an older rootfs that was newest back then).

While I still think that this is a feature best supported by znapzend directly (possibly, with additional znapzendzetup schedule option to at all consider cloning/promotion for `rpool/ROOT` and `rpool/zones` type of source datasets), I also realize that this is something difficult to get right even without the perl complexities :-D and it was easier for me to at least PoC the logic in shell. After all, it can be ported to main tool later, once we know the chain of renames, clones, promotes and sends actually works. (Alas, some of this may be dependent on ZFS version and its capabilities as well -- this was used on recent illumos, so older Solaris or different Linux mileage may vary).

So at this point the script, while still quite experimental (and something to be called at user's understanding, discretion and risk), is a bit closer to production in that I can configure its activities better by feeding it envvars with dataset names and other options, and it can diagnose whether the situation is a "simple" one it can handle (one "history owner" on src and dst to promote) or if humans should take a look and pre-clean the sandbox by moving away confusing items.

I have used it already to rectify my backups of OS rootfs and of local zones, one at a time, and it was able to handle cloning and promotion of the "history owner" dataset on destination, as well as cloning of other new (not yet backed up) datasets that have it as origin. It can also "iterate" to clone from other origins (children of history owner) then - sometimes the OS history was not linear.

So until the next upgrade, I can use the usual `znapzend` to backup safely; though after such upgrade I should call this script a few times first to promote datasets on my backup disk, and branch off datasets comprising my obsoleted rootfs as clones there - and then re-enable znapzend again.

So, hoping it helps directly, and indirectly by landing ideas into main code eventually, and yet washing my hands in case it explodes... :-D


oetiker (approved):

oetiker:
thanks :) any feedback from people using it ?

jimklimov:
So far just myself AFAIK - it helped (with a fair amount of manual work to
launch for many certain system and zone rootfses) to streamline my OI
workstation updates to let the laptop continue being a regular znapzend
citizen. :)

PR #553, opened 2021-07-29
contrib/synczbe: refactor work with zoned property + improve behavior for recently cloned sources

jimklimov:
A bit of refinement during/after recent OI upgrade :)

For full disclosure: in this use-case I ran `pkg` rituals to create a new boot environment and update it, and local zones too; then I ran `beadm activate` to have that new BE be the one used next boot (in particular, this `zfs promote`'s its rootfs and zone roots), and ran this script against my system and zone roots to apply similar change to the backup pool, e.g.:

````
DEBUG=no SR=nvpool/zones/testdhcp/ROOT ~jim/shared/znapzend/contrib/synczbe

DEBUG=no ZFSLIST_REGEX=firefly SR=nvpool/ROOT ~jim/shared/znapzend/contrib/synczbe
DEBUG=no ZFSLIST_REGEX=hipster SR=nvpool/ROOT ~jim/shared/znapzend/contrib/synczbe
````

Subsequent `znapzend` runs for standard snapshot and resync use-case seems to have worked decently (no new non-cloned datasets spawned) before and after the OI reboot.

As before, `zoned` property is optionally (ab)used on my local backup dataset to surely avoid automounting of the backed-up datasets, with the overhead that it complicates non-trivial manipulations with those datasets.

### DIFF
--- a/contrib/synczbe
+++ b/contrib/synczbe
@@ -88,6 +88,7 @@ optionalDisarmDestZonedProp() {
     if [ "$DR_ZBE_ZONED_VALUE" = on ]; then
 
         if ! $DR_ZONED_DISARMED && [ "$DR_ZONED_VALUE" = on ]; then
+            echo "Disarming 'zoned' property on destination '$DR' ..."
             DR_ZONED_DISARMED=true
             $ZFSW set zoned=off "$DR"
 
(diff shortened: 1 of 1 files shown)


jimklimov:
gentle bump  :)

oetiker (dismissed):

oetiker (approved):

oetiker:
ready to merge ?

jimklimov:
I guess so - there were few things to address over my recent OS upgrades
(and related dataset promotions on the backup).

Next big thing might be to add the smarts into znapzend itself by porting
this logic well-tested standalone (to detect if target needs a
promotion/clone like this and to perform it). Also the script here acts on
a (zbe or rootfs) hosting dataset specified by caller, which for me means a
rootfs/ROOT and a dozen local zone roots listed in another script to call
this one.

One practical caveat - better stop znapzend before upgrading the OS to
avoid confusion and avoid it sending a full copy of newly discovered
dataset(s) to backup, then `beadm activate NEWBE` if update went well, then
use synczbe here to update the backup, and only then `init 6` to reboot
into new OS revision - with that uptime starting znapzend service in usable
conditions again.

PR #581, opened 2022-08-01
Fix VERSION replacement perl command in Makefile

drscream:
Remove space from search and replace commend for the VERSION number.
Based on the code in the bin-folder no space should be required after
the VERSION comment.

See also:
https://github.com/oetiker/znapzend/blob/master/bin/znapzend#L10

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -54,7 +54,7 @@ dist_bin_SCRIPTS = $(BIN)
 docs: $(POD) $(MAN)
 
 dist-hook: docs
-	$(PERL) -i -p -e 's/^my .*#VERSION/my \$$VERSION = q{$(PACKAGE_VERSION)}; #VERSION/' $(distdir)/$(BIN)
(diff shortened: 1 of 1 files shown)


Makefile.am
```diff
@@ -54,7 +54,7 @@ dist_bin_SCRIPTS = $(BIN)
 docs: $(POD) $(MAN)
 
 dist-hook: docs
-	$(PERL) -i -p -e 's/^my .*# VERSION/my \$$VERSION = q{$(PACKAGE_VERSION)}; # VERSION/' $(distdir)/$(BIN)
+	$(PERL) -i -p -e 's/^my .*#VERSION/my \$$VERSION = q{$(PACKAGE_VERSION)}; #VERSION/' $(distdir)/$(BIN)
```
oetiker, line 57:
```suggestion
	$(PERL) -i -p -e 's/^my .*#\s+VERSION/my \$$VERSION = q{$(PACKAGE_VERSION)}; # VERSION/' $(distdir)/$(BIN)
```

oetiker (approved):

PR #584, opened 2022-08-15
Mojo::IOLoop::Delay is no longer needed

mtelka:


jimklimov:
Curious, so it was used (in Perl terms) but not actually used (referenced by code) anymore?

mtelka:
> Curious, so it was used (in Perl terms) but not actually used (referenced by code) anymore?

I think `Mojo::IOLoop::Delay` was never used explicitly.  It was used implicitly by `Mojo::IOLoop::ForkCall` only.

The `use Mojo::IOLoop::Delay` line was added to `ZFS.pm` in Jan 2022 via https://github.com/oetiker/znapzend/commit/492bd16e53a78d893c6f950aefde3a765c09ca7d to solve some warning.  But then, in May 2022, the (deprecated) `Mojo::IOLoop::ForkCall` was replaced by `Mojo::IOLoop::Subprocess` - see https://github.com/oetiker/znapzend/commit/0a656f75c64cf36a96c1bddb20fa9c6bd50de16b - so `Mojo::IOLoop::Delay` is no longer used/needed at all.

oetiker (approved):
indeed

Issue #585, opened 2022-08-24
Use of uninitialized value $ssh[11] in open at /usr/lib/ZnapZend/ZFS.pm line 212


jimklimov:
In my `dmesg` I see a series of messages in SUBJ, grouped as a burst of 10 and another 7 half a minute later.

Current codebase has it a bit shifted by line numbers, but otherwise this area seems same as installed:

https://github.com/oetiker/znapzend/blob/27f965ff84fc496d5d2c93bd74b2a745666633fa/lib/ZnapZend/ZFS.pm#L209-L215

This is in `listSnapshots`, and looking at debug output and the code, I believe it is trying to parse a dataset name with colons (the only thing that apparently differs from successful logs) and `splitHostDataSet` fails to do it properly (can colons be treated as e.g. port separator for general-case SSH remote?) Colons in dataset names are injected by current OpenIndiana `pkg` handling of backup rootfs'es.

https://github.com/oetiker/znapzend/blob/27f965ff84fc496d5d2c93bd74b2a745666633fa/lib/ZnapZend/ZFS.pm#L40-L42

Afterwards per symptoms, `$dataSet` is `undef` and operations with it emit warnings (two actually, just one in dmesg though) until it fails in call to `zfs` program where it specifies no dataset to list:

````
[2022-08-24 18:11:18.11843] [29435] [debug] checking to clean up snapshots of source rpool/zones/nut/jenkins-worker/ROOT/openindiana-2022:03:15-backup-1
Use of uninitialized value $ssh[11] in join or string at /usr/lib/ZnapZend/ZFS.pm line 211.
# zfs list -H -o name -t snapshot -s creation -d 1
Use of uninitialized value $ssh[11] in open at /usr/lib/ZnapZend/ZFS.pm line 212.
cannot open '': empty component in name
[2022-08-24 18:11:18.12961] [29435] [debug] got nothing to clean in source rpool/zones/nut/jenkins-worker/ROOT/openindiana-2022:03:15-backup-1
````

jimklimov:
Looking at the regex, it seems like trying to catch the port is indeed the culprit (for the local path, without any hostspec in the string)

jimklimov:
[...]

PR #588, opened 2022-10-21
Document and facilitate use of Restricted shell (on SSH receiver end)

jimklimov:
Some lessons learned from the setup made to `znapzend` from cloud-hosted servers to a remote SSH server.
If someone gets to those cloud servers, I don't want them to nuke *all* of that backup box too ;)

### DIFF
--- a/lib/ZnapZend/ZFS.pm
+++ b/lib/ZnapZend/ZFS.pm
@@ -1431,10 +1431,32 @@ sub fileExistsAndExec {
     my $remote;
 
     ($remote, $filePath) = $splitHostDataSet->($filePath);
-    my @ssh = $self->$buildRemote($remote, [@{$self->priv}, qw(test -x), $filePath]);
+    my @ssh1 = $self->$buildRemote($remote, [@{$self->priv}, qw(test -x), $filePath]);
+    # Note: with restricted shell setup, users may be unable to run
+    # fully qualified command names (with a slash) but may run what
+    # they have in the PATH.
+    my @ssh2 = $self->$buildRemote($remote, [@{$self->priv}, qw(command -v), $filePath]);
 
-    print STDERR '# ' . join(' ', @ssh) . "\n" if $self->debug;
-    return !system(@ssh);
+    print STDERR '# ' . join(' ', @ssh1) . " || " . join(' ', @ssh2) . "\n" if $self->debug;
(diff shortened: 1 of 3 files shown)


jimklimov:
NOTE: I wonder if `znapzendzetup --help` should make it clearer that `DST` is not the place to pass advanced SSH arguments (port etc.) as I stumbled upon while looking at #585?

lib/ZnapZend/ZFS.pm
```diff
@@ -1430,10 +1430,32 @@ sub fileExistsAndExec {
...
+
+    print STDERR '# ' . join(' ', @ssh1) . " || " . join(' ', @ssh2) . "\n" if $self->debug;
+    # Tricks to hide output of `command -v` from perl stdout:
+    if (!system(@ssh1)) {
+        return 1;
+    }
 
-    print STDERR '# ' . join(' ', @ssh) . "\n" if $self->debug;
-    return !system(@ssh);
+    my $ssh2out = `@ssh2`;
```
oetiker, line 1445: hmmm this will invoke a shell locally, we try hard not todo that in the rest of the code ...

oetiker (dismissed):

lib/ZnapZend/ZFS.pm
jimklimov: Well, normally (in systems where existing code worked) it would not get to this line and `test-x` with `!system()` would be good enough, local or remote.

So this line and below only strike in situations where current code would assume `filePath` is absent (e.g. no `mbuffer` to run).

Is there a particular reason to avoid local shell?

Alternately, is there a good way to have similar effect (probably over SSH and with quirky setup of restricted shell) other than pulling in complex IPC3 or something of the sort?

oetiker: the problem is that shells have a tendency to be scrambled by user configuration and also to do unexpected things with the input ... the fewer shells the fewer problems ... we already got one at the other end of the ssh connection

jimklimov: But here we rely exactly on POSIX (IIRC) built-in `command` and profile settings like `PATH`...

oetiker: lets try :)

Issue #591, opened 2022-10-21
Seems that for non-enabled mode we only delete snaps from the first child dataset


jimklimov:
Support for disabling snapshots of certain children goes as recursive snapshot of the parent and removal of these snaps for certain child datasets. In practice I see however only the first one cleaned up this way (both in `oracleMode` and without):

````
[2022-10-21 16:25:42.49053] [4980] [info] creating recursive snapshot on rpool/export
# zfs snapshot -r rpool/export@znapzend-auto-2022-10-21T16:25:42Z
[2022-10-21 16:25:43.05007] [4980] [info] Requesting removal of marked datasets: rpool/export/home/abuild/.ccache@znapzend-auto-2022-10-21T16:25:42Z, rpool/export/home/abuild/jenkins-nut@znapzend-auto-2022-10-21T16:25:42Z, rpool/export/home/abuild/jenkins-nut-doc@znapzend-auto-2022-10-21T16:25:42Z
# zfs destroy -r rpool/export/home/abuild/.ccache@znapzend-auto-2022-10-21T16:25:42Z
[2022-10-21 16:25:43.50717] [2915] [debug] snapshot worker for rpool/export done (4980)
[2022-10-21 16:25:43.51451] [2915] [debug] send/receive worker for rpool/export spawned (5215)
[2022-10-21 16:25:43.51698] [5215] [info] starting work on backupSet rpool/export
# zfs list -H -r -o name -t filesystem,volume rpool/export
...
````

and that's it - the only `destroy` in that run's log.

jimklimov:
So, somehow this bit misfires:
````
# lib/ZnapZend.pm
createSnapshot = { ...
         $self->zZfs->destroySnapshots(@dataSetsExplicitlyDisabled, 0);

# lib/ZnapZend/ZFS.pm
sub destroySnapshots {
     my @toDestroy = ref($_[0]) eq "ARRAY" ? @{$_[0]} : ($_[0]);
     my @recursive = $_[1] ? ('-r') : ();
...
````

Actually the `$_` contains a long array with each item from original `dataSetsExplicitlyDisabled`

I was among last to touch this code in 0bdfece7 actually, but built upon earlier existing spelling with `@`-varname. Other usages of the method pass a dollar-varname like:
````
$self->zZfs->destroySnapshots($toDestroy, 1);
````

jimklimov:
Yep, that was it - REFERENCE to array `\@` did it. Now there's some bug inside the destroy method :)

````
[...]

PR #592, opened 2022-10-22
Fix deletion of many snaps for many datasets, and handle several not-"enabled" sub-trees under one schedule

jimklimov:
Closes: #591


oetiker:
thank you!

Issue #589, opened 2022-10-21
PR self-tests fail in docker stage: `mbuffer (no such package)`


jimklimov:
e.g. https://github.com/oetiker/znapzend/actions/runs/3297260330/jobs/5437862941

````
...
(33/33) Installing zfs (0.8.3-r1)
Executing busybox-1.31.1-r11.trigger
Executing ca-certificates-20191127-r2.trigger
OK: 67 MiB in 47 packages
fetch http://dl-3.alpinelinux.org/alpine/edge/community/x86_64/APKINDEX.tar.gz
fetch http://dl-cdn.alpinelinux.org/alpine/v3.11/main/x86_64/APKINDEX.tar.gz
WARNING: Ignoring http://dl-3.alpinelinux.org/alpine/edge/community/: UNTRUSTED signature
fetch http://dl-cdn.alpinelinux.org/alpine/v3.11/community/x86_64/APKINDEX.tar.gz
ERROR: unable to select packages:
  mbuffer (no such package):
    required by: world[mbuffer]
The command '/bin/sh -c apk add --no-cache zfs curl bash autoconf automake nano perl=${PERL_VERSION} openssh &&   apk add --no-cache --repository http://dl-3.alpinelinux.org/alpine/edge/community/ mbuffer &&   ln -s /dev/stdout /var/log/syslog &&   ln -s /usr/bin/perl /usr/local/bin/perl' returned a non-zero code: 1
Error: Process completed with exit code 1.
````

jimklimov:
Empirically, this seems to only run for the first build in a PR. If I later iterate and add commits, only the 3 perl version tests run and result is green :)

jimklimov:
So far failed to fix but can reproduce: with Alpine 3.11 it seems the trusted keys have expired vs. "community" repo (with mbuffer). Whether I use newer alpine 3.16 for newer key trust, or ignore security issues, it fails to install mbuffer because:
````
...
#6 4.130 OK: 70 MiB in 50 packages
#6 4.187 /lib/libcrypto.so.1.1
#6 4.296 /lib/libcrypto.so.3
#6 4.296 /usr/lib/libcrypto.so.1.1
#6 4.296 /usr/lib/libcrypto.so.3
#6 4.300 fetch http://dl-cdn.alpinelinux.org/alpine/edge/community/x86_64/APKINDEX.tar.gz
[...]

PR #593, opened 2022-10-22
Add .editorconfig and .gitattributes

jimklimov:
Effectively force LF in checkouts, as hit during issue #589 exploration and building with Docker for Windows (so `COPY` carries over "wrong" EOLs).


oetiker (approved):
great

PR #619, opened 2024-01-08
Fix parser for root (only) dataset names - investigate

jimklimov:
Follows up from #585 and a Gitter discussion https://matrix.to/#/!XZJhhFzueFpkcSXLHR:gitter.im/$gVcghYo80LIz2zIdyJQyfxRYJkyb07mQjoB5bGH5zTU?via=gitter.im&via=inf.ethz.ch&via=matrix.org (and later posts during the day)

The crux of it is that `znapzend` v0.21.2 (current release) got confused when a backup schedule was defined on a root dataset of a pool AND the custom `tsformat` included colons to separate hours-minutes-seconds:

````
[2024-01-08 09:57:54.87309] [154415] [debug] === getDataSetProperties():
        Collected: $VAR1 = {
          'pre_znap_cmd' => 'off',
          'mbuffer' => 'off',
          'enabled' => 'on',
          'src_plan' => '1months=>1weeks,1years=>1months,10years=>6months',
          'mbuffer_size' => '128M',
          'zend_delay' => '0',
          'post_znap_cmd' => 'off',
          'src' => 'bpool',
          'dst_dst_0_plan' => '1months=>1week,1years=>1month,10years=>6months',
          'recursive' => 'on',
          'tsformat' => 'znapzend-auto-%Y-%m-%dT%H:%M:%SZ',
          'dst_dst_0' => 'znapzend:pond/export/DUMP/ci-deb/bpool'
        };

...

[2024-01-08 09:57:54.89115] [154583] [info] creating recursive snapshot on bpool
# ssh -o batchMode=yes -o ConnectTimeout=30 bpool@znapzend-auto-2024-01-08T10 zfs snapshot -r '22:13Z'
ssh: Could not resolve hostname znapzend-auto-2024-01-08t10: Name or service not known
# ssh -o batchMode=yes -o ConnectTimeout=30 22 zfs list -H -o name -t snapshot 13Z
````

So a request to snapshot a local `bpool` dataset as `bpool@znapzend-auto-2024-01-08T10:22:13Z` got interpreted by the generic routine as `user=bpool`, `host=znapzend-auto-2024-01-08T10`, `dsname=22:13Z` and an absent snap. The latter part also got further parsed for the subsequent command into `host=22` and `dsname=13Z`, it seems.

This visibly strikes in `sub createSnapshot {...}`, `sub destroySnapshots {...}` and probably other consumers of `sub $splitHostDataSet`.

If the "DST" definition remains (per help/man) as `[[user@]host:]dataset` where `dataset` is fixed for DST but may be `dsname[@snap]` for general parsing (and `snap` may have some but not all sorts of funny characters) -- we don't really have good criteria (for regex or beside it) to tell apart a `user@host` from partial `dataset@snap` strings, in some cases, it seems.

There are various ideas in that thread that can be pursued as separate PRs. This one lays the foundations for such pursuits, by adding a few run-time sanity checks (e.g. to avoid destructive actions with bogus values), and some self-test code to gauge success of different solution attempts. A large helper in this effort is the `t/znapzend-lib-splitter.t` script which calls whatever implementations we have in `ZFS.pm` and runs them against a matrix of known remote, dataset and snapname strings (concatenated into what can be seen in production from configs and ZFS queries), to see if they get parsed back properly.

### CODE REVIEW

File: lib/ZnapZend/ZFS.pm [hunk 1]
```diff
@@ -54,7 +54,12 @@ my $splitHostDataSet = sub {
 };
 
 my $splitDataSetSnapshot = sub {
-    return ($_[0] =~ /^([^\@]+)\@([^\@]+)$/);
+    my $count = ($_[0] =~ tr/@//);
```
oetiker, line 57: This, I think, will actually modify the $_[0] with this the argument given when calling the function, this is unwanted action at a distance I think. I would suggest to work with a copy

oetiker: `my $snap  = shift;`

oetiker (approved):

PR #620, opened 2024-01-08
Tame the failing spellcheck job

jimklimov:
Do not check some files, like AUTHORS, the generated man pages and CHANGES.old (and its commit hash strings).
Add domain-specific words to whitelist.

Also try to fix self-testing environment in the docker container.

PR #621, opened 2024-01-08
Fix parser for root (only) dataset names

jimklimov:
Includes PRs #619 (see for most context of the initial problem) and #620 (so also addresses spellchecker passing after these changes).

After some tinkering with the splitter logic vs. conjured-up test cases, just one case remained where we can only define our preference (at least, lacking some other method arguments or wholly a new class to convey the information explicitly): how to treat `X@Y:Z` strings - are they a remote `user@host:rootds` or a local `rootds@funny:snapname`?

So the self-test defines this one use-case as a "bogus" one with somewhat reasonably "hard-coded" specific expectations:

> NOTE: In absence of any hints we can not reliably discern below
> a     task='poolrootfs@snap-2:3'
> vs.   task='username@hostname:poolrootfs'
> which one is a local pool's root dataset with a funny but legal
> snapshot name, and which one is a remote user@host spec with a
> remote pool's root dataset. For practical purposes, we proclaim
> preference for the former: we are more likely to look at funny
> local snapshot names, than to back up to (or otherwise care about)
> remote pools' ROOT datasets.

...and so the self-tests for the parser are enabled as part of standard test runs here.

Also this uncovered deficiencies in `ZnapZend::Config` variant of the parser which got fixed now by the regex proposed in #585 (and improved upon here for `ZnapZend::ZFS`). It does not have to care about snapshots, so is a lot simpler.


oetiker:
how about assuming case user@host:destds unless user actually is a local rootds ?

jimklimov:
> > how about assuming case user@host:destds unless user actually is a local rootds ?

>  For practical purposes, we proclaim preference for the former: we are more likely to look at funny local snapshot names, than to back up to (or otherwise care about) remote pools' ROOT datasets.

Yep, so we declare that `user@host:destds` means local pool named `user` (and its root dataset), and its snapshot `host:destds`.

oetiker:
what I mean is that the code could actually test(!) if a given string actually is a local root dataset

jimklimov:
Ah, must have misread your earlier comment. Yes, I've had such thoughts too - just did not get to implementing them :)

Instead, wondered whether:

* reliability:
  * do we have some context that such pool does not exist currently so we can not reliably check if it is a pool?
    * e.g. a pluggable HDD/flash seems likely; rotation of several normally-vaulted media devices is a good practice for redundancy
    * pre-emptive setup to create a pool (more so via znapzend) seems unlikely
  * can then check if the `host` part can be resolved, but for systems with intermittent networking (VPN?) this may also not be bullet-proof
* a chicken-and-egg problem:
  * our generic ZFS calls IIRC use this helper method to decide if they should go over SSH :)
  * should we go for custom call-out to a local `zfs` tool and/or special tweaks to the common method?

oetiker:
you are right this might be fragile ... another idea is this:

we know the actual src and destination configurations ... they do not include snapshots, so at that time things are still clear ... 

so I guess we just need to alter the parsing flow

jimklimov:
> I guess we just need to alter the parsing flow

As noted earlier, for the least intrusive change, I am currently inclined in favor of further arguments (caller knows if the context deals with possible snapshots or strictly "live" dataset names).

Not sure if such refinement has to be part of this PR, or if you deem this one as an improvement (less-broken state of codebase) and it can be merged as a stepping-block to someone (else) making and diligently testing that change. Coding the change does not feel that hard, but testing and chizeling would likely need more time than I could currently share.

jimklimov:
It may also be worthwhile to use `ZnapZend::Config` definitions of the splitting method for strings coming from the config, as opposed to those generally juggled  by `ZnapZend::ZFS`. OOP-wise, these strings reflect sort of different object classes. But other than such an idea, I have not much more about implementing something of the sort.

oetiker (approved):

oetiker (changes requested):
please update the CHANGES file

jimklimov:
Cheers, is the update okay? :)

oetiker (approved):

.github/workflows/spelling/expect.txt
```diff
@@ -383,6 +383,7 @@ notest
 NOTMAKE
 nroff
 NSTIPS
+nsub
```
Ariessanea1, line 386: Submit

Ariessanea1: > Submit

.github/workflows/spelling/expect.txt
```diff
@@ -383,6 +383,7 @@ notest
...
 nvpool
```
Ariessanea1, line 387: Pool

lib/ZnapZend/Config.pm
```diff
@@ -47,7 +47,9 @@ has backupSets => sub { [] };
 
 ### private functions ###
 my $splitHostDataSet = sub {
```
Ariessanea1, line 49: Set host bate

t/znapzend-lib-splitter.t
```diff
@@ -4,7 +4,7 @@
 # since there are many use-cases and combinations to take care of.
 # We do so below by constructing "task" strings from components we
 # know to be a dataset name and some defined (or not) remote spec
-# and/or snapshot name, and deconstructing it back with the class
+# and/or snapshot name, and de-constructing it back with the class
```
Ariessanea1, line 7: Congratulations

.github/workflows/spelling/expect.txt
```diff
@@ -383,6 +383,7 @@ notest
 NOTMAKE
 nroff
 NSTIPS
```
Ariessanea1, line 385: Stops

.github/workflows/spelling/expect.txt
```diff
@@ -419,6 +420,7 @@ plaintar
 plist
 plugin
 png
```
Ariessanea1, line 422: Phnom

.github/workflows/spelling/expect.txt
```diff
@@ -461,6 +463,7 @@ rmdir
 rmprog
 rmtmp
 ron
+rootds
```
Ariessanea1, line 466: Roots

lib/ZnapZend/Config.pm
Ariessanea1, line 49: Subslithostbateset

jimklimov: @Ariessanea1 : what's with the spam across many repos?

.github/workflows/spelling/expect.txt
Ariessanea1: Sund

Ariessanea1 (commented):
- [ ] Things m

Ariessanea1 (commented):
- [ ] Things m

Ariessanea1 (commented):
- [ ] Things m

Issue #622, opened 2024-01-15
CI: Problem building Docker image for "Release/Docker push" scenario: repo lacks mbuffer?


jimklimov:
Apparently the chosen distro/repo lacks `mbuffer`:

* https://github.com/oetiker/znapzend/actions/runs/7514122856/job/20456645200
````
------
 > [runtime 2/4] RUN   apk add --no-cache zfs curl bash autoconf automake nano perl=5.30.3-r0 openssh &&   apk add --no-cache --repository http://dl-3.alpinelinux.org/alpine/edge/community/ mbuffer &&   ln -s /dev/stdout /var/log/syslog &&   ln -s /usr/bin/perl /usr/local/bin/perl:
1.612 Executing busybox-1.31.1-r11.trigger
1.617 Executing ca-certificates-20191127-r2.trigger
1.665 OK: 67 MiB in 47 packages
1.691 fetch http://dl-3.alpinelinux.org/alpine/edge/community/x86_64/APKINDEX.tar.gz
2.211 WARNING: Ignoring http://dl-3.alpinelinux.org/alpine/edge/community/: UNTRUSTED signature
2.211 fetch http://dl-cdn.alpinelinux.org/alpine/v3.11/main/x86_64/APKINDEX.tar.gz
2.371 fetch http://dl-cdn.alpinelinux.org/alpine/v3.11/community/x86_64/APKINDEX.tar.gz
2.559 ERROR: unable to select packages:
2.569   mbuffer (no such package):
2.569     required by: world[mbuffer]
------
Dockerfile:22
--------------------
````

jimklimov:
OOps, got closed by PR mis-posted to my repo :)

PR #624, opened 2024-01-15
Dockerfile: bump Alpine Linux version

jimklimov:
Following https://stackoverflow.com/a/73374817/4715872 suggestions: Alpine signing key was rotated, so new package builds are not trusted by the old image

Closes: #622

NOTE: Whatever builds I do, it only runs "Testing in image" but not "Build runtime image" stage. So can't really tell if this fix works :\ Spun up Docker Desktop locally, but it is "loading metadata" for the past 6 minutes, so I don't know if it would proceed further.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -41,7 +41,8 @@ CMD [ "znapzend --logto=/dev/stdout" ]
 FROM builder as test
 
 RUN \
-  cpan Devel::Cover
+  cpan Devel::Cover && \
+  cpan Test::SharedFork
 
 RUN \
   ./test.sh
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,7 +1,7 @@
 
 ##### Builder image
 ARG PERL_BUILD_VERSION=5.30-buster
-ARG ALPINE_VERSION=3.11
+ARG ALPINE_VERSION=3.19
 FROM docker.io/library/perl:${PERL_BUILD_VERSION} as builder
 
 WORKDIR /usr/local/src


oetiker (approved):

jimklimov:
Hm, on upstream/master builds it is still upset - now with perl version (we still reference a probably too-old one):

````
#10 [runtime 2/4] RUN   apk add --no-cache zfs curl bash autoconf automake nano perl=5.30.3-r0 openssh &&   apk add --no-cache --repository http://dl-3.alpinelinux.org/alpine/edge/community/ mbuffer &&   ln -s /dev/stdout /var/log/syslog &&   ln -s /usr/bin/perl /usr/local/bin/perl
#10 0.794 fetch https://dl-cdn.alpinelinux.org/alpine/v3.19/main/x86_64/APKINDEX.tar.gz
#10 1.004 fetch https://dl-cdn.alpinelinux.org/alpine/v3.19/community/x86_64/APKINDEX.tar.gz
#10 1.458 ERROR: unable to select packages:
#10 1.462   perl-5.38.2-r0:
#10 1.462     breaks: world[perl=5.30.3-r0]
#10 1.462     satisfies: autoconf-2.71-r2[perl]
#10 1.462                automake-1.16.5-r2[perl]
#10 ERROR: process "/bin/sh -c apk add --no-cache zfs curl bash autoconf automake nano perl=${PERL_VERSION} openssh &&   apk add --no-cache --repository http://dl-3.alpinelinux.org/alpine/edge/community/ mbuffer &&   ln -s /dev/stdout /var/log/syslog &&   ln -s /usr/bin/perl /usr/local/bin/perl" did not complete successfully: exit code: 3
------
````

But at least it seems 5.38.x is a suggested way to go - will post an updated PR :)

PR #626, opened 2024-01-15
Handle not-sending of not-enabled datasets

jimklimov:
Follows up from ideas in #623

UPDATE: There was a thought about using `oracleMode` to fall back from a recursive snapshot send to one-by-one processing in case we have exclusions (and/or send recursively only some smaller sub-trees that are safe to send - no exclusions there), but looking at this code structure I do not see a recursive *send* in fact. Maybe it is hidden deeper in `$self->zZFS->...`?..


jimklimov:
> aren't you changeing the same function as in the other PR ?

As noted in the other PR - indeed, same function but for a different purpose. The change here merely skips sending of datasets that are "not-enabled" by whatever definition of that is in the codebase. That other PR changes the definition.

lib/ZnapZend.pm
```diff
@@ -600,8 +600,13 @@ my $sendRecvCleanup = sub {
             my $dstDataSet = $srcDataSet;
             $dstDataSet =~ s/^\Q$backupSet->{src}\E/$backupSet->{$dst}/;
 
-            $self->zLog->debug('sending snapshots from ' . $srcDataSet . ' to ' . $dstDataSet . ((grep (/^\Q$srcDataSet\E$/, @dataSetsExplicitlyDisabled)) ? ": not enabled, should be skipped" : ""));
-            {
+            my $srcDataSetDisabled = (grep (/^\Q$srcDataSet\E$/, @dataSetsExplicitlyDisabled));
+            $self->zLog->debug('sending snapshots from ' . $srcDataSet . ' to ' . $dstDataSet . ($srcDataSetDisabled ? ": not enabled, should be skipped" : ""));
```
oetiker, line 604: you are skipping, way "should" ?

oetiker (changes requested):

lib/ZnapZend.pm
jimklimov: IIRC, copy-pasted from other skip messages, so kept in same style: note what we should do by hints from the world, eventually do take some step forward.

jimklimov: Ok, wording addressed. 

Now I wonder if this message should only be seen `... if $self->debug` - configuration disablement is rather permanent and presumably intentional, so this has a chance to be regular and not particularly useful spam in daemon logs (journal, dmesg), unlike the earlier warnings about skips due to run-time hiccups (comms to backup server, inconsistent snapshot history, etc.).

jimklimov: Or maybe not. We still post the snapshots we DO send, so this is just another line in the wall of text.

oetiker (approved):

PR #631, opened 2024-01-17
Streamline init for FreeBSD

jimklimov:
Import and adapt https://cgit.freebsd.org/ports/blame/sysutils/znapzend/files/znapzend.in?id=e5474f0f70649c210a347a94c9c0bcf1124c2952 to simplify the (still manual) installation routine into FreeBSD (and relatives) from source

Routine placed into `init/README.md` was checked on a FreeBSD 12 VM, service works via `/etc/rc.d/znapzend` and auto-starts on reboot; also honours the `/etc/defaults/znapzend` options.


oetiker (changes requested):
please update the CHANGES file

oetiker (approved):

PR #636, opened 2024-02-08
Extend `--autoCreation` effect (or default lack thereof) to newly appearing sub-datasets

jimklimov:
For example, avoid unexpected creation of avoidably large replicas of root datasets where promotable clones of boot environments are used (currently `znapzend` makes a new full `zfs send`, not a clone+promote operation - see #503 and others after it).

Also added a `--noautoCreation` option to help override configuration file settings (where used), primarily to help test this change locally.

NOTE: I have a nagging feeling that the default value for this setting belongs in ZFS properties (per-destination) rather than in CLI as one toggle to rule them all (can be kept for one-off overrides). Maybe another PR would address that, more so if I get to solving #503 directly. UPDATE: See #637 for zfs props approach.

### DIFF
--- a/.github/workflows/spelling/expect.txt
+++ b/.github/workflows/spelling/expect.txt
@@ -374,6 +374,7 @@ newestbe
 nf
 nh
 noaction
+noauto
 nobase
 nodelay
 nodestroy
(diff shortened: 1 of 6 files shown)


oetiker (dismissed):

oetiker:
I think the zfs aproach #637 is sensible ... do you see benefit in this still ?

jimklimov:
Somewhat yes. Properties allow for default settings, CLI options for run-time decisions. And this PR allows the latter to be consistent when nested datasets are involved :)

oetiker (approved):

oetiker (approved):

Issue #651, opened 2024-05-27
`make DESTDIR=... install` of `znapzend` also sends CPANM modules to that location when building from scratch


jimklimov:
Found while investigating for issue #646:
````
:; make distclean ; rm -rf /dev/shm/zzz-0.22.0 ; ./bootstrap.sh ; ./configure ; make DESTDIR=/dev/shm/zzz-0.22.0 install &&  find /dev/shm/zzz-0.22.0 -ls
````

ends up with a gazillion files like `/dev/shm/zzz-0.22.0/home/jim/znapzend/thirdparty/lib/perl5/Test/Mojo.pm` (offset by `DESTDIR` and my `HOME`-based `THIRDPARTY_DIR` value).

Will post a fix for that as well.

Issue #646, opened 2024-05-02
make install does not respect LIBDIR, always installs to LIBDIR/../lib/


spike77453:
Since znapzend 0.22.0 `make install` does not respect `LIBDIR` any more and tries to install the `*.pm` files to `LIBDIR/.../lib/`:

<details>
  <summary><pre>$ ./configure --libdir=/usr/share/perl5/vendor_perl/</pre></summary>

```console
checking in to see how you are doing... keep fighting man!
checking for a BSD-compatible install... /usr/bin/install -c
checking whether build environment is sane... yes
checking for a race-free mkdir -p... /usr/bin/mkdir -p
checking for gawk... gawk
checking whether make sets $(MAKE)... yes
checking whether make supports nested variables... yes
checking whether UID '0' is supported by ustar format... yes
checking whether GID '0' is supported by ustar format... yes
checking how to create a ustar tar archive... gnutar
checking whether to enable maintainer-specific portions of Makefiles... no
checking whether make supports nested variables... (cached) yes
checking for perl... /usr/bin/perl
checking for curl... /usr/bin/curl
checking for wget... /usr/bin/wget
checking for pod2man... /usr/bin/pod2man
checking for rm... /usr/bin/rm
checking for CONFIG_YEAR and CONFIG_DATE... got 2024 and 2024-05-03 respectively
checking default SMF service name... oep/znapzend
checking for way to cat URLs... /usr/bin/wget -O -
checking for perl version greater than or equal to 5.16.0... checking is perl reasonably complete... yes. ExtUtils::MakeMaker is available
checking if require a c compiler to get perl modules compiled... no
checking for grep that handles long lines and -e... /usr/bin/grep
checking for gnumake... no
checking for gmake... /usr/bin/gmake
checking for gnu make availability... /usr/bin/gmake is GNU make
checking the price for bergulian eckels... way to expensive!
checking for PERL_MODULES... $(top_srcdir)/lib/ZnapZend.pm $(top_srcdir)/lib/ZnapZend/Config.pm $(top_srcdir)/lib/ZnapZend/Time.pm $(top_srcdir)/lib/ZnapZend/ZFS.pm $(top_srcdir)/lib/inheritLevels.pm 
checking for PERL_CONFIG_VERSION... 5.38
checking for PERL_THIRDPARTY_DIST cache... empty
configure: Generating "data" files from templates, see below for executable scripts
configure: Generating templated script files that should be marked executable
checking that generated files are newer than configure... done
[...]

PR #652, opened 2024-05-27
Fix `make install` of Perl modules with a custom `DESTDIR` and/or `libdir`

jimklimov:
Closes: #646
Closes: #651

Checked fixed with GNU Make on Ubuntu 22, illumos make on OpenIndiana, BSD make on FreeBSD 12.

See "screenshot" at https://github.com/oetiker/znapzend/issues/646#issuecomment-2133971445

### DIFF
--- a/.github/workflows/spelling/expect.txt
+++ b/.github/workflows/spelling/expect.txt
@@ -360,6 +360,7 @@ missingpool
 Mkbootstrap
 mkdir
 mkdirprog
+Mksymlists
 modled
 mojolicious
 Morbo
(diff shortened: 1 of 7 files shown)


oetiker (dismissed):

oetiker:
thanks!

oetiker (approved):

jimklimov:
Right, thanks for the `CHANGES` :)

jimklimov:
For historic note about this fix: it seems that `make DESTDIR=... install` as the first and only `make` invocation in the use-case I had here could have been the culprit about not succeeding with `cpanm`/`carton` preparation phases, at least greater than parallelism. The former is easy to prove experimentally, the latter less so :)

PR #655, opened 2024-05-28
Rename `lib/inheritLevels.pm` to `lib/ZnapZend/InheritLevels.pm`…

jimklimov:
…to minimize confusion

From issue #646 discussion follow-up

Needs diligent testing (CI and practical) to make sure the expected `inherit` functionality is still handled correctly. For local testing to progress, this effort led to PR #654 first, and existing tests did catch some syntax issues that were fixed over a few iterations of this PR's preparation.

### DIFF
--- a/lib/ZnapZend/InheritLevels.pm
+++ b/lib/ZnapZend/InheritLevels.pm
@@ -1,5 +1,6 @@
-package inheritLevels;
+package ZnapZend::InheritLevels;
 # Note: classes made by Class::Struct may not be sub-namespaced
+# (UPDATE: comment might be valid at least in older Perl versions?)
 
 use Class::Struct;
 
@@ -10,7 +11,7 @@ use Class::Struct;
 ### For "usual" datasets (filesystem,volume) `zfs` returns the
 ### properties inherited from higher level datasets; but for
 ### snapshots it only returns the same - not from higher snaps.
-struct ('inheritLevels' => {
+struct ('ZnapZend::InheritLevels' => {
     zfs_local => '$', # set to ask for properties defined locally in dataset
     zfs_inherit => '$', # set to ask for properties inherited from higher datasets
     zfs_received => '$', # set to ask for properties received during "zfs send|recv" (no-op for now, mentioned for completeness)
@@ -59,9 +60,9 @@ sub getInhMode {
 sub reset {
     # Without args, just resets all fields back to undef so they can be
     # re-assigned later. With an arg tries to import a legacy setting by
-    # number or string, or copy from another instance of inheritLevels.
+    # number or string, or copy from another instance of InheritLevels.
     my $self = shift;
-    
+
     $self->zfs_local(undef);
     $self->zfs_inherit(undef);
     $self->zfs_received(undef);
@@ -69,7 +70,7 @@ sub reset {
     if (@_) {
         my $arg = shift;
         # Assign from legacy values
-        if ($arg->isa('inheritLevels')) {
+        if ($arg->isa('ZnapZend::InheritLevels') or $arg->isa('InheritLevels')) {
             $self->zfs_local($arg->zfs_local);
             $self->zfs_inherit($arg->zfs_inherit);
             $self->zfs_received($arg->zfs_received);
@@ -89,7 +90,7 @@ sub reset {
         } elsif (!defined($arg) or $arg == undef) {
             1; # No-op, keep fields undef
         } else {
-            warn "inheritLevels::reset() got unsupported argument '$arg'\n";
+            warn "ZnapZend::InheritLevels::reset() got unsupported argument '$arg'\n";
             return 0;
         }
     }
@@ -102,14 +103,14 @@ __END__
 
 =head1 NAME
 
-inheritLevels - helper struct for various options of ZFS property inheritance
+ZnapZend::InheritLevels - helper struct for various options of ZFS property inheritance
 
 =head1 SYNOPSIS
 
-use inheritLevels;
+use ZnapZend::InheritLevels;
 use ZnapZend::ZFS;
 ...
-my $inherit = new inheritLevels;
+my $inherit = new ZnapZend::InheritLevels;
 $inherit->zfs_local(1);
 $inherit->zfs_inherit(1);
 my $properties = $self->getSnapshotProperties($snapshot, $recurse, $inherit, @dstSyncedProps);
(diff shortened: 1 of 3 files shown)


lib/ZnapZend/InheritLevels.pm
```diff
@@ -102,14 +103,14 @@ __END__
...
+ZnapZend::InheritLevels - helper struct for various options of ZFS property inheritance
 
 =head1 SYNOPSIS
 
-use inheritLevels;
+use ZnapZend::InheritLevels;
 use ZnapZend::ZFS;
 ...
-my $inherit = new inheritLevels;
+my $inherit = new ZnapZend::InheritLevels;
```
oetiker, line 113:
```suggestion
my $inherit = ZnapZend::InheritLevels->new;
```

lib/ZnapZend/ZFS.pm
```diff
@@ -547,13 +547,13 @@ sub mostRecentCommonSnapshot {
...
     # See definition of recognized $inherit values for this context in
-    # getSnapshotProperties() code and struct inheritLevels.
+    # getSnapshotProperties() code and struct ZnapZend::InheritLevels.
     my $inherit = shift; # May be not passed => undef
     if (!defined($inherit)) {
         # We leave defined but invalid values of $inherit to
         # getSnapshotProperties() to figure out and complain,
         # but for this routine's purposes set a specific default.
-        $inherit = new inheritLevels;
+        $inherit = new ZnapZend::InheritLevels;
```
oetiker, line 556:
```suggestion
        $inherit = ZnapZend::InheritLevels->new;
```

lib/ZnapZend/ZFS.pm
```diff
@@ -1321,13 +1321,13 @@ sub getSnapshotProperties {
     }
 
     if (!defined($inherit)) {
-        $inherit = new inheritLevels;
+        $inherit = new ZnapZend::InheritLevels;
```
oetiker, line 1324:
```suggestion
        $inherit = ZnapZend::InheritLevels->new;
```

lib/ZnapZend/ZFS.pm
```diff
@@ -1321,13 +1321,13 @@ sub getSnapshotProperties {
...
         $inherit->zfs_local(1);
     } else {
         # Data type check
-        if ( ! $inherit->isa('inheritLevels') ) {
-            $self->zLog->warn("getSnapshotProperties(): inherit argument is not an instance of struct inheritLevels");
-            my $newInherit = new inheritLevels;
+        if ( ! ($inherit->isa('ZnapZend::InheritLevels') or $inherit->isa('InheritLevels')) ) {
+            $self->zLog->warn("getSnapshotProperties(): inherit argument is not an instance of struct ZnapZend::InheritLevels");
+            my $newInherit = new ZnapZend::InheritLevels;
```
oetiker, line 1330:
```suggestion
            my $newInherit = ZnapZend::InheritLevels->new;
```

lib/ZnapZend/ZFS.pm
```diff
@@ -1459,7 +1459,7 @@ sub getSnapshotProperties {
                 # Go up to root of the pool, without recursing into other children
                 # of the parent datasets/snapshots, and without inheriting stuff
                 # that is not locally defined properties of a parent (or its parent).
-                my $inherit_local_recurseparent = new inheritLevels;
+                my $inherit_local_recurseparent = new ZnapZend::InheritLevels;
```
oetiker, line 1462:
```suggestion
                my $inherit_local_recurseparent = ZnapZend::InheritLevels->new;
```

oetiker (changes requested):
I think `new Class` syntax is discurraged

jimklimov:
> I think `new Class` syntax is discouraged

I had a suspicion about that - more so after debugging the problems which arose (lack of new method reported). From the times "back when" I remember I've wondered if "class" and "struct" are internally same (syntactic sugar) or different - I think this syntax was copied from some example about structs when I converted from a map 1a925f61 to a struct 8f1f3810 and eventually moved it to a separate file in df23fabe.

And probably it should go with parentheses? e.g. `->new()` not plain `->new;` (at least per examples I see in other places of the script).

oetiker:
> > I think `new Class` syntax is discouraged
> 
> I had a suspicion about that - more so after debugging the problems which arose (lack of new method reported). From the times "back when" I remember I've wondered if "class" and "struct" are internally same (syntactic sugar) or different - I think this syntax was copied from some example about structs when I converted from a map [1a925f6](https://github.com/oetiker/znapzend/commit/1a925f61d1c87db845f1245da3d0857757fbcb95) to a struct [8f1f381](https://github.com/oetiker/znapzend/commit/8f1f38102ebdd8eb91327a35b488ad1b019c5278) and eventually moved it to a separate file in [df23fab](https://github.com/oetiker/znapzend/commit/df23fabeffc12dc4c74e53a5a5cb3ad33aae9e4e).
> 
> And probably it should go with parentheses? e.g. `->new()` not plain `->new;` (at least per examples I see in other places of the script).

the brackets in perl are optional since functions and variables can be distinguished by the presence of a sigil

lib/ZnapZend/ZFS.pm
```diff
@@ -1321,13 +1321,13 @@ sub getSnapshotProperties {
...
     if (!defined($inherit)) {
-        $inherit = new inheritLevels;
+        $inherit = ZnapZend::InheritLevels->new;
         $inherit->zfs_local(1);
     } else {
         # Data type check
-        if ( ! $inherit->isa('inheritLevels') ) {
-            $self->zLog->warn("getSnapshotProperties(): inherit argument is not an instance of struct inheritLevels");
-            my $newInherit = new inheritLevels;
+        if ( ! ($inherit->isa('ZnapZend::InheritLevels') or $inherit->isa('InheritLevels')) ) {
```
oetiker, line 1328: why do you allow both ? can there still be InheritLevels

oetiker (commented):
see comment

lib/ZnapZend/ZFS.pm
jimklimov: Good catch. Was not sure which sort of name it reacts to, so while developing - kept both to make sure the change is viable at all :\

jimklimov:
Rebased over recent master and trying the latest suggestion.

jimklimov:
At least, survived the tests :)

oetiker (approved):

PR #684, opened 2025-11-11
Update README.md for spelling and Debian packages.

bill-mcgonigle:
fixes Gregy/znapzend-debian#23 .


oetiker:
writing things with z is intentional :) you would find many more instances of that all throughout the package

PR #686, opened 2026-03-18
Handle Perl 5.42 stricter strict refs by avoiding deref'ing scalar 0.

sjuxax:
We get output like this otherwise:

Broken on Perl 5.42:
```
znapzend on  fix/perl-5.42-strictrefs [!?] via 🐪 v5.42.0
at 22:25:36 ❯ znapzendzetup edit zs/work
Can't use string ("0") as an ARRAY ref while "strict refs" in use at /usr/bin/znapzendzetup line 261.
```

Fixed:
```
znapzend on  fix/perl-5.42-strictrefs [!?] via 🐪 v5.42.0
at 22:26:09 ❯ bin/znapzendzetup edit zs/work
ERROR: cannot get backup config
```


bin/znapzendzetup
```diff
@@ -258,7 +258,9 @@ sub main {
         if ($#ARGV == grep { /^--(?:$ignoreOpts)$/ } @ARGV){
             $opts->{src} = pop @ARGV;
 
-            my $backupSet = $zConfig->getBackupSet(0, 0, $opts->{src})->[0]
+            my $backupSets = $zConfig->getBackupSet(0, 0, $opts->{src})
+                or die "ERROR: cannot get backup config\n";
+            my $backupSet = $backupSets->[0]
                 or die "ERROR: cannot get backup config\n";
```
oetiker, line 264: the second case can never be true. please drop it too.

PR #689, opened 2026-04-18
Developer headache: generated files tracked in Git

jimklimov:
Various files provided or generated by autotools, as well as man pages generated from POD sections, have little business being in Git. They just complicate `git add -p` etc. workflows.

This PR drops them from git and adds the names to `Makefile.am` so the files made available on the build system by running `./bootstrap.sh` (in whichever autotools version it boasts) are added to `make dist`, and revises `.gitignore` files.

Note there are also files whose contents are apparently re-touched due to `VERSION` bumps in the newest entry of the `CHANGES` text (`README.md`, `COPYRIGHT`, `bin/*`), but that at least makes sense.

Checked with an out-of-tree build and test using just the resulting tarball, that nothing pops up in `git status` now.
* Using `make distcheck` was initially problematic due to distribution of `bin/cpanm` in the tarball - I fixed that to pass, but am not sure the tool SHOULD be in the tarball (also depends on external providers, and involves `carton` and half a dozen other scripts which we do not distribute anyway).


jimklimov:
Spellchecker had a bunch of complaints, some about a ton of keywords that it deems not needed in `expect.txt` - added a commit which drops those; another which re-orders the `excludes.txt` - both prepared via command from GHA failure suggestion:
```
:; curl -s -S -L 'https://raw.githubusercontent.com/check-spelling/check-spelling/main/apply.pl' \
    | perl - 'https://github.com/jimklimov/znapzend/actions/runs/24606203997/attempts/1'
```

Oddly this check is only seen on a developer fork (e.g. in listing of Branches), not on the upstream (e.g. as part of PR).

UPDATE: Should get fixed by this PR now - so PRs also do spell-checking just in time for contributors to notice and take up the chore.

oetiker (approved):

PR #693, opened 2026-05-06
Drop POD rendered docs from Git

jimklimov:
They are generated for tarball during build.

Follows-up from #689, including its CHANGES entry.


hadfl:
> They are generated for tarball during build.
> 
> Follows-up from #689, including its CHANGES entry.

I think these are added so it's possible to read the rendered docs on GitHub.

jimklimov:
Fair point, I didn't realize that but GH does render them. Maybe they should be generated (so force-synced with latest `*.pm` sources, not best-effort like now) and pushed as a GH website?

PR #694, opened 2026-05-06
bin/znapzendzetup: document the `--recursive` option in POD section

jimklimov:
subj :)

### DIFF
--- a/.github/workflows/spelling/expect.txt
+++ b/.github/workflows/spelling/expect.txt
@@ -49,6 +49,7 @@ buildpackage
 Bzip
 canmount
 CBuilder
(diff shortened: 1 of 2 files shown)


oetiker:
@copilot resolve the merge conflicts in this pull request

jimklimov:
got the comment... now, will look into it

jimklimov:
I see, the conflict is due to #704 dropping the spellcheck altogether. This one was implemented in a very annoying manner, but I wonder if the concept had to be dropped? In NUT we have decent results with `aspell` which can be tested locally as part of `make check`.

PR #696, opened 2026-06-01
Fix mostRecentCommonSnapshot discarding the live common snapshot

moetiker:
A list slice left over from an earlier refactor causes `mostRecentCommonSnapshot()` to always discard the live result of `lastAndCommonSnapshots()`.

## The bug

`lastAndCommonSnapshots()` returns a three-element list `($lastSnapshot, $lastCommonSnapshot, $dstSnapCount)`. The call applied a `[1]` list slice to that return value:

```perl
($lastSnapshot, $lastCommonSnapshot, $dstSnapCount) = ($self->lastAndCommonSnapshots(...))[1];
```

The slice selects only the second element and assigns it to `$lastSnapshot`, leaving `$lastCommonSnapshot` (and `$dstSnapCount`) `undef`. Because `$lastCommonSnapshot` is always undef, the routine always falls through to the property-based fallback (`dst_X_synced`), which is meant only as a fallback for *offline* destinations and can return a snapshot that no longer exists on the destination.

## Origin

- 2018 (`8a262a1`): original form `$lastCommonSnapshot = (...)[1];` — correct, slice picks index 1 into a single scalar.
- 2020 (`9aee085`, *"fix use of lastAndCommonSnapshots()"*): the left-hand side was widened to the three-element list to capture all return values, but the `[1]` slice was not removed — silently breaking the assignment.

The fallback masked this for online destinations (synced properties usually match the live state), which is why it went unnoticed.

## Why this matters (safety / data-loss relevance)

The value returned by `mostRecentCommonSnapshot()` is the snapshot that `sendRecvCleanup()` **protects from source retention cleanup**:

```perl
my $recentCommon = $self->zZfs->mostRecentCommonSnapshot(...);
if ($recentCommon) {
    @{$toDestroy} = grep { $recentCommon ne $_ } @{$toDestroy};   # protect it
}
```

That snapshot is the base required for the next incremental `zfs send -I`. With the bug, when a destination is **online** but its saved `dst_X_synced` property is stale (e.g. points at a source-only snapshot that is not actually on the destination), the function protects the wrong snapshot — leaving the **genuine common base eligible for deletion**. Destroying that base breaks incremental replication: it forces a full re-send, or fails outright under `--forbidDestRollback`.

The fix makes the live (authoritative) common snapshot win, so the snapshot that actually exists on both sides is the one protected.

Scope of the behaviour change:
- **Offline destinations** (the `--cleanOffline` path): `lastAndCommonSnapshots` finds no live common, so both old and new code use the property fallback — **behaviour is unchanged** here.
- **Normal steady state** (online, sync succeeded): live common == the just-synced newest snapshot == what the property points at — **no observable difference**.
- The fix only changes the edge case "online destination + stale/partial property", where it is strictly safer.

## The fix

Drop the slice so all three values are captured, restoring the live common snapshot as the primary source with the property lookup as the intended fallback.

## Tests

`t/znapzend-lib-most-recent-common.t` (registered in `test.sh`) sets up, via the mock `zfs`, the exact risk scenario: source and destination share the *oldest* snapshot, while the `dst_X_synced` property marks the *newest* (destination-absent) snapshot as synced. It asserts:

1. the live common snapshot is returned (not the property guess);
2. the returned snapshot is **not** the stale-property (newest source-only) one;
3. the returned snapshot is the genuine common base that exists on the destination;
4. **emulating the production protection idiom, the real common base is excluded from the destroy list** (i.e. protected from cleanup), while source-only snapshots remain eligible.

Verified RED→GREEN: with the bug reintroduced, the assertions for (1), (3) and especially (4) fail — the real base ends up in the destroy list and would be deleted; with the fix all pass. The full existing suite stays green.


jimklimov:
For cross-reference, commits listed above correspond to PRs:
* #384 => #506
* #507

I guess I played a role in both instances, especially the later PR with introduction of that mistake, sorry. I suppose I'll do something like that again and again, nice to have automated reviewers that know better than ourselves how to actually implement the vague ideas we have :P

PR #698, opened 2026-06-01
Fix spurious leftover-snapshot warning (array ref in scalar context)

moetiker:
In the `allowDestRollback == 2` resync path of `sendRecvSnapshots`, after destroying the filter-matched (znapzend-made) snapshots on the destination, the code re-counts what remains in order to warn the admin about *manually-named* snapshots that would make `zfs recv -F` fail:

```perl
$self->destroySnapshots($self->listSnapshots($dstDataSet, $snapFilter, undef));
...
my $dstSnapCountAll = scalar($self->listSnapshots($dstDataSet, qr/.*/, undef));
...
$self->zLog->warn("ERROR: some snapshot(s) not covered by znapzend filter still exist on destination ...")
    if (!$lastCommon && $dstSnapCountAll>0);
```

`listSnapshots` returns an **array reference**. `scalar($ref)` is always `1`, so `$dstSnapCountAll > 0` is always true. The alarming "some snapshot(s) not covered by znapzend filter still exist … the zfs send+receive would likely fail below!" warning is therefore emitted **even in the normal case**, where the destination held only znapzend-made snapshots and is now clean after the destroy above.

## Fix

Dereference the array so the count reflects the snapshots actually remaining:

```perl
my $dstSnapCountAll = scalar(@{$self->listSnapshots($dstDataSet, qr/.*/, undef)});
```

## Testing

This branch is reached only through the `--sinceForced` promotion logic (`$doPromote > 1`) feeding `allowDestRollback == 2` into `sendRecvSnapshots`, with a destination that has snapshots but no common one — a deep, narrow path whose only effect here is a log message. A faithful automated test would require reconstructing that whole `--since` promotion scenario plus log capture, which is disproportionate; the change is a self-contained scalar-context correction (same class as several `scalar(@{...})` uses already in the file). The full existing test suite was run and stays green.


jimklimov:
Probably originally my error, facepalming myself just in case and thanking you for the fix :)

